PR #305, opened 2018-12-21
amended post build events to ease installer creation

### LINKED ISSUE #304: Grasshopper_UI - Post build events for the installer

epignatelli:
Fixes #304 

When building the Grasshopper_Toolkit we are now creating a copy of the AppData\Roaming\Grasshopper\Libraries\BHoM into AppData\Roaming\BHoM\InstallFiles\Grasshopper

This is to ensure consistencies when building the installer and avoid copying and pasting flying files from who-knows-from-which-folder-to-choose

### DIFF
--- a/Grasshopper_UI/Grasshopper_UI.csproj
+++ b/Grasshopper_UI/Grasshopper_UI.csproj
@@ -363,9 +363,14 @@
   </Target>
   -->
   <PropertyGroup>
-    <PostBuildEvent>call "$(TargetDir)UI_PostBuild.exe" ..\..\ "C:\Users\$(Username)\AppData\Roaming\Grasshopper\Libraries\BHoM"
+    <PostBuildEvent>rename "$(TargetName).dll" "$(TargetName).gha"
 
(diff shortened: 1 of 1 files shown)


epignatelli:
Yeah, I agree definitely. I was planning to apply this post build only on Release mode when you compile from VS, but I didn't have the time.

So debug, triggers developer mode copy of the files, and Release triggers the copy for the installation.

epignatelli:
Hi guys, happy to move this feature to the CodeAssistant in some other form.

Can I ask to add this as a priority?
I did this pull request because it has been an absolute pain to create the files for the installer, and we had no slick procedure setup. I just want to avoid doing it again, because it's very unreliable, and if done wrong cancels all the job we do.

Happy to have a call if you know already what to do and don't have time, I can pick it up!
I'll close this!

awakeman:
The idea is for the continuous integration system that @adecler is working on to build the installer(s) and that a human should never need to. We're looking at having that working as a prototype by the end of next week. The code server will have all the dependencies in place to build the full stack so shouldn't be an issue.

PR #320, opened 2019-01-18
Fixes #312 - Do not reset 2.0 components if they fail to deserialise

### LINKED ISSUE #312: Grasshopper_UI: v2.0 component resets if it fails to deserialize

epignatelli:

Closes #312

v2.0 components run into a type not found exception, because of major changes in the BHoM and other toolkits. Preventing the exception by moving the m_Method set operations into the null guard and adding a try-catch for the creation of the type. The `ToType` method is only used by the v2.0 components.

### Test files

You can use the unit tests to check if the errors persists. Below is a good example (try it before and after switching to this branch). You can also make use of any other script that contains old UI components to test this.
https://burohappold.sharepoint.com/:u:/s/BHoM/EYH3MBD5ihdJrBwkb6OHTWoBGJjCIIvHJvLfT7CGmVBlsQ?e=q55wQ6

### Definition of Done
If the components do not reset themselves (turn red, no input, no outputs, and get thrown at the origin of the grasshopper canvas), the fix is successful.

### DIFF
--- a/Grasshopper_UI/2.0/Templates/MethodCallTemplate.cs
+++ b/Grasshopper_UI/2.0/Templates/MethodCallTemplate.cs
@@ -293,13 +293,13 @@ namespace BH.UI.Grasshopper.Templates
                 Type type = typeString.ToType();
                 RestoreMethod(type, methodName, paramTypes);
 
-                m_IsDeprecated |= m_Method.IsDeprecated();
-                m_IsNotImplemented = m_Method.IsNotImplemented();
-                m_IsReleased = m_Method.IsReleased();
 
                 // Restore the ports
                 if (m_Method != null)
                 {
+                    m_IsDeprecated |= m_Method.IsDeprecated();
+                    m_IsNotImplemented = m_Method.IsNotImplemented();
+                    m_IsReleased = m_Method.IsReleased();
                     Type outputType = (m_Method is MethodInfo) ? ((MethodInfo)m_Method).ReturnType : m_Method.DeclaringType;
                     ComputeDaGets(m_Method.GetParameters().ToList(), outputType);
                 }
--- a/Grasshopper_Engine/Convert/ToType.cs
+++ b/Grasshopper_Engine/Convert/ToType.cs
@@ -59,7 +59,17 @@ namespace BH.Engine.Grasshopper
             if (parts.Length == 0)
                 return null;
             else
-                return BH.Engine.Reflection.Create.Type(parts[0]);
+            {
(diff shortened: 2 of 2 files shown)


adecler:
The code you have pushed here makes sense in a general term but I don't see any difference on the example you provide. Both cases are keeping inputs and outputs intact and giving me this (which seems expected):

image

Grasshopper_UI/2.0/Templates/MethodCallTemplate.cs
```diff
@@ -245,15 +250,38 @@ protected override void SolveInstance(IGH_DataAccess DA)
...
+            }
+
+            if (m_StoredMethodName != null)
+            {
+                writer.SetString("MethodName", m_StoredMethodName);
+            }
+            else if (!isMethodNull)
+            {
                 writer.SetString("MethodName", m_Method.Name);
-                writer.SetInt32("NbParams", parameters.Count());
```
IsakNaslundBh: This parameter is no longer serialised. Needed.

Grasshopper_UI/2.0/Templates/MethodCallTemplate.cs
```diff
@@ -282,24 +315,29 @@ public override bool Read(GH_IO.Serialization.GH_IReader reader)
                         m_IsDeprecated = true;
                     }
 
-                    paramTypes.Add(paramType.ToType());
+                    Type parameterType = paramType.ToType();
+
+                    if (parameterType == null)
+                    {
+                        m_LoadingWarnings.Add($"The type {paramType} cannot be found. Type is set to System.Object");
```
IsakNaslundBh: Message should say that the type is set to null rather than System.Object?

Grasshopper_UI/2.0/Templates/MethodCallTemplate.cs
```diff
@@ -188,13 +188,18 @@ protected override void SolveInstance(IGH_DataAccess DA)
 
             BH.Engine.Reflection.Compute.ClearCurrentEvents();
 
+            foreach (string warn in m_LoadingWarnings)
+            {
+                Engine.Reflection.Compute.RecordWarning(warn);
```
IsakNaslundBh: Would be good some more information about what this actually means. Current simple message will be very cryptic for most users.

IsakNaslundBh:
Think this works as expected now for old scripts. Helpful that it tries to find a method as well as it can. Makes it easier to check old scripts.

PR #321, opened 2019-01-23
Grasshopper ui issue314 update component correctly explode

### LINKED ISSUE #314: Grasshopper_Toolkit: Grasshopper objects are modified during a solution
It seems that we now have an error message every time we delete a (group of) component(s) that is feeding into others:

image

epignatelli:

Closes #314 

Ok, this is the start of tweaking the Grasshopper UI to more explicitly handle the dialogue between Grasshopper itself and the BHoM_UI.
It is fixing the point where all the error were exposed: the explode component. I definitively disabled the autoupdates. The component now expires if the input changed. It auto-explode only the first time you connect an input.
Moreover, the base CallerComponent now does not expire the solution, but the display, that forces the parameter to update. CreateType now explicitly expires the solution.

#### Changed:
- from `public virtual void CallerComponent.RefreshComponent()` to `public virtual CallerComponent.OnItemSelected(object sender = null, object e = null)`

#### Removed:
- `private void ExplodeComponent.UpdateOutputs(bool ignoreAutoUpdate)`

#### Added:
- `public virtual void CallerComponent.OnBHoMUpdates(object sender = null, object e = null)`
- `public virtual void CallerComponent.OnGrasshopperUpdates(object sender = null, object e = null)`
- `public override void ExplodeComponent.OnGrasshopperUpdates(object sender = null, object e = null)`
- `public override void ExplodeComponent.OnGrasshopperUpdates(object sender = null, object e = null)`
- `private bool ExplodeComponent.IsExpired()`

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/EaADcbbmXdJKsYYDFyzsRE0Bfe2yR4sXOYp420lb8RuZDg?e=7zogsP

### What to check
I'd rather not condition the way you test this. A general UX feedback, especially about what is annoying would be perfect.

### Additional comments

FYI @al-fisher @rwemay

### DIFF
--- a/Grasshopper_UI/2.1/Components/Engine/Explode.cs
+++ b/Grasshopper_UI/2.1/Components/Engine/Explode.cs
@@ -76,16 +76,8 @@ namespace BH.UI.Grasshopper.Components
         {
             this.OnGrasshopperUpdates();
 
-            if (this.IsExpired())
-            {
-                Engine.Reflection.Compute.ClearCurrentEvents();
-                Engine.Reflection.Compute.RecordWarning("Output paramters do not match object properties. Please right click and <Update Outputs>");
-                Logging.ShowEvents(this, BH.Engine.Reflection.Query.CurrentEvents());
-            }
-            else
-            {
+            if (!this.IsExpired())
                 base.SolveInstance(DA);
-            }
         }
 
         /*******************************************/
@@ -93,8 +85,24 @@ namespace BH.UI.Grasshopper.Components
         protected override void AfterSolveInstance()
         {
             if (Caller.OutputParams.Count != 0 & Params.Output.Count == 0)
+            {
                 this.OnBHoMUpdates();
-            base.BeforeSolveInstance();
+            }
+            else if (this.IsExpired())
+            {
+                if (Params.Output.All(p => p.Recipients.Count == 0))
+                {
+                    this.OnBHoMUpdates();
+                }
+                else
+                {
+                    Engine.Reflection.Compute.ClearCurrentEvents();
+                    Engine.Reflection.Compute.RecordWarning("Output paramters do not match object properties. Please right click and <Update Outputs>");
+                    Logging.ShowEvents(this, BH.Engine.Reflection.Query.CurrentEvents());
+                }
+            }
+
+            base.AfterSolveInstance();
         }
 
         /*******************************************/
--- a/Grasshopper_UI/2.1/Components/Engine/Explode.cs
+++ b/Grasshopper_UI/2.1/Components/Engine/Explode.cs
@@ -47,7 +47,7 @@ namespace BH.UI.Grasshopper.Components
         /**** Public Override Methods           ****/
         /*******************************************/
 
-        public override void OnBHoMUpdates(object sender = null, object e = null)
+        public void OnBHoMUpdates(object sender = null, object e = null)
         {
             RecordUndoEvent("OnBHoMUpdates");
             // Forces the component to update
@@ -58,7 +58,7 @@ namespace BH.UI.Grasshopper.Components
 
         /*******************************************/
 
-        public override void OnGrasshopperUpdates(object sender = null, object e = null)
+        public void OnGrasshopperUpdates(object sender = null, object e = null)
         {
             // Update the output params based on input data
             Params.Input[0].CollectData();
--- a/Grasshopper_UI/2.1/Components/Engine/Explode.cs
+++ b/Grasshopper_UI/2.1/Components/Engine/Explode.cs
@@ -42,8 +42,6 @@ namespace BH.UI.Grasshopper.Components
 
         public override Caller Caller { get; } = new ExplodeCaller();
 
-        public bool AutoUpdateOutputs { get; set; } = true;
-
 
         /*******************************************/
         /**** Constructors                      ****/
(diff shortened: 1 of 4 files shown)


Grasshopper_UI/2.1/Templates/CallerComponent.cs
```diff
@@ -104,13 +104,17 @@ public virtual void RefreshComponent()
...
-            if (document != null) // this prevents expiring a solution while another solution is running
-            {
-                ExpireSolution(true);
-            }
+            this.OnDisplayExpired(true);
         }
 
+        /*******************************************/
+
+        public virtual void OnBHoMUpdates(object sender = null, object e = null) { }
```
adecler: Does it have to be here? I don't see any use of it except in the Explode component.
Same for `OnGrasshopperUpdates`.

Grasshopper_UI/2.1/Templates/CallerComponent.cs
```diff
@@ -104,13 +104,17 @@ public virtual void RefreshComponent()
             this.RegisterOutputParams(null); // We call its bits individually: input, output 
             this.Params.OnParametersChanged(); // and ask to update the layout with OnParametersChanged()
 
-            GH_Document document = OnPingDocument(); // when a solution is running the document of the component is set to null
-            if (document != null) // this prevents expiring a solution while another solution is running
-            {
-                ExpireSolution(true);
-            }
+            this.OnDisplayExpired(true);
```
adecler: This is causing new problems and I am not sure it actually fixes anything. Can you give an example where using ExpireSolution would be bad here? Here's one where using OnDisplayExpire is causing trouble:

expiresolutionproblem

Grasshopper_UI/2.1/Components/oM/CreateType.cs
```diff
@@ -44,15 +44,16 @@ public class CreateTypeComponent : CallerComponent
...
         {
-            base.RefreshComponent();
+            base.OnItemSelected();
 
             Type type = Caller.SelectedItem as Type;
             if (type != null)
                 Message = type.ToText();
-        }
 
+            ExpireSolution(true);
```
adecler: Might not be necessary. See comment below.

Grasshopper_UI/2.1/Templates/CallerComponent.cs
epignatelli: No, definitely they are not doing anything. The only reason I explicitly put them there is to treat the CallerComponent as an interface, where these methods are part of it. With the benefit of not forcing child classes to overwrite them. 
The idea is "I am in the ExplodeComponent, I have these two methods at my service I can use to solve the Grasshopper-BHoM dialogue".
But that's not a big deal, it is more about clarity. Happy to remove them if you think it the game is not worth the candle!

epignatelli:
Sorry - I definitely missed a key point: I didn't want to completely remove the autoupdate. I was planning to re-add it properly, but I prioritized making the explode work and do not break every time. I can keep working on this pr if we want to do it now.

Now, the main reasons why I am against the auto update of the outputs are:
- If you spent 15 mins wiring all the outputs to something else and, by mistake, you change the input, you lose your work (there is no CTRL + Z) for that. We basically delete people's efforts.
- The solution needs to run in order for the explode to know what to set as ouputs. Even if we do register the outputs on the solution end, we can't guarantee that the explode is the leaf of the flow graph, so part of it won't be calculated. Worse if we have a chain of that. Viceversa, if there is something else that depends on it, we can't set the outputs, because the solution is still running.

I also agree on the behavior above, we should let people use the explode the same way we use a panel for different objects, just to see. I can definitely come up with something for that!

Grasshopper_UI/2.1/Templates/CallerComponent.cs
epignatelli: Sorry - In general, the issue raises from CallerComponent expiring the solution + Explode that autoupdates

IsakNaslundBh:
I am with @epignatelli regarding the auto update. To me it is one of those things that feel like a good idea at first, but really isn't, as it has the potential of fully messing up your scripts.

I am a strong advocate for the behaviour just implemented; Auto update first time, then the user have to make a conscious decision. Even this manual update I suppose could lead to wires being lost, but at least it happens by a user action.

Grasshopper_UI/2.1/Templates/CallerComponent.cs
adecler: I think it is just better to keep things simple. If something is only used in one place, don't give the illusion it is part of a larger picture.

Grasshopper_UI/2.1/Templates/CallerComponent.cs
```diff
@@ -104,13 +104,24 @@ public virtual void RefreshComponent()
...
-            GH_Document document = OnPingDocument(); // when a solution is running the document of the component is set to null
-            if (document != null) // this prevents expiring a solution while another solution is running
+            if (Caller.InputParams.Count == 0 | Caller.InputParams.All(p => p.HasDefaultValue))
             {
-                ExpireSolution(true);
+                this.ExpireSolution(true);
+            }
+            else
+            {
+                this.OnDisplayExpired(true);
```
adecler: I am still confused by this one but that might just be because I haven't worked on the GH code for a while. 
It looks like you are trying really hard to limit the cases where the output of the component is updated BUT this method is only called when the user has selected the item for this component. In that situation you will always want to update your output, no? You seems to acknowledge that yourself by overwriting the method for CreateTypeComponent which is the only component where the user can select an item multiple times.

Grasshopper_UI/2.1/Templates/CallerComponent.cs
adecler: see comment below

adecler:
@IsakNaslundBh , yeah agreed. I keep changing my mind back and forth between auto-update and not. I feel the answer is something in between. What about this?

- The explode component will automatically update its outputs when you reconnect its input. This means it only happens when the user is doing a conscious action on the component. 
- Do not auto-update when the data in the wire changed but the wire has not been touched. In this case, just turn the component to orange and offer the manual update.  
- Implement the support for the Ctrl-Z functionality if possible

This way, I am trying to keep the expected behaviour for most scenarii:
- new explode component, input plugged for teh first time: auto-update outputs
- copy/paste of the component to explode different data: auto-update outputs
- change input connection of the explode component to look at other data: auto-update outputs
- any other case: manual update only

I have my doubts about the input changed situation so....
Alternatively, a more conservative solution would be to do this:
- if there is no output connections: auto-update (we don't risk to loose connection anyway)
- if there is: only manual update is allow to make sure we don't break stuff.

`No output connectors` definitely cover the two most important cases for auto-update: new component and copied/pasted component. So I would be happy with that.

IsakNaslundBh:
Like the last suggestion by @adecler. Only update the outputs of the explode when there are no outputs connected. A clear rule from all sides, and something that can be easily explained and should give the best benefits of both sides of things.

Grasshopper_UI/2.1/Templates/CallerComponent.cs
epignatelli: No, I agree, that's probably just me being overcautious about this.
I think you want to update the outputs just if there is something to calculate. 
If we morph a `Create` into a `Create.Node(...)` we wouldn't update, because you just selected the item and you have no inputs to run the method. Unless the component either has no inputs or all of them have a default value.

I am definitely not attached to this solution :D I just thought it was safer, but I trust more you if you think there issues with it.

epignatelli:
I appreciate guys all the feedback, thanks!

I think the two solutions @adecler proposed are one the increment of the other.
I pushed the updates, enabling the behaviour for the conservative one, and can definitely explore the braver one later on. It will be just a matter of enabling the feature at each "WireChanged" event (which doesn't exist, we might need to resort to `ObjectChanged` or something similar).

Grasshopper_UI/2.1/Templates/CallerComponent.cs
adecler: Sorry, I am going to be a pain on this one. 
To summarise what your code is doing here, the solution will always be expired unless, somehow, the user had a chance to provide an input. 
The only case where I could find it happening is if you create a custom object, create an input, plug something in, and then select its type. 
Unfortunately, your code is doing exactly the wrong thing here since the output would still be a custom object despite its type selected to something else. Here's what I mean:

image

I think this is only possible as a scenario because the custom object component is still broken but this makes an important point: adding unnecessary code is always dangerous because, while it serves no purpose, it can potential create edge effects on future code you didn't anticipate.  

Also, please always use `||` instead of `|`. This is especially important when the first condition is protecting against an exception on the evaluation of the second but a good practice for all situations.

Grasshopper_UI/2.1/Components/Engine/Explode.cs
```diff
@@ -42,42 +42,74 @@ public class ExplodeComponent : CallerComponent
...
         }
 
         /*******************************************/
 
-        protected override void RegisterInputParams(GH_InputParamManager pManager)
+        protected override void AfterSolveInstance()
         {
-            if (Params.Input.Count == 0)
-                base.RegisterInputParams(pManager);
+            if (Caller.OutputParams.Count != 0 & Params.Output.Count == 0)
```
adecler: Please always use `&&` instead of `&`. This is especially important when the first condition is protecting against an exception on the evaluation of the second but a good practice for all situations.

Grasshopper_UI/2.1/Components/Engine/Explode.cs
```diff
@@ -86,33 +118,28 @@ protected override void AppendAdditionalComponentMenuItems(ToolStripDropDown men
...
 
-        private void RefreshLabel_Click(object sender, EventArgs e)
+        private bool IsExpired()
         {
-            ExpireSolution(false);
-            UpdateOutputs(true);
-        }
-
-        /*******************************************/
+            if (Caller.OutputParams.Count == 0 | Params.Output.Count == 0)
```
adecler: Please always use `||` instead of `|`. This is especially important when the first condition is protecting against an exception on the evaluation of the second but a good practice for all situations.

Grasshopper_UI/2.1/Components/Engine/Explode.cs
```diff
@@ -86,33 +118,28 @@ protected override void AppendAdditionalComponentMenuItems(ToolStripDropDown men
...
+            {
+                return true;
+            }
 
-        private void UpdateOutputs(bool ignoreAutoUpdate)
-        {
-            bool doJob = ignoreAutoUpdate | AutoUpdateOutputs;
```
adecler: Please always use `||` instead of `|`. This is especially important when the first condition is protecting against an exception on the evaluation of the second but a good practice for all situations.

al-fisher:
@adecler - see my review above. Happy changes satisfy your comments. I will merge

PR #329, opened 2019-02-06
Grasshopper toolkit issue308 create custom forced type

### LINKED ISSUE #302: Grasshopper_Toolkit: CreateCustom - property name not updating until you add another property
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

### LINKED ISSUE #308: Grasshopper_Toolkit: CreateCustom - populating the inputs by selecting an existing type doesn't work anymore
I am referring to this:

image

When doing this, the name of the component changes but the inputs are not added.

NB: be careful not to break the serialisation of this component when fixing the problem.

epignatelli:

Closes #308 
Closes #302 
This pr is:
- Reinstating the possibility of forcing a type when creating a custom, and making the Grasshopper component respond to that. 
- Updating the component when a parameter name changes 
- While checking that the `TypeHint` for IBHoMGeometry is working I noticed another issue that required fix: the `FromGoo` method now always checks if the type is a Rhino type and converts to BHoM if necessary. In the current master branch it was failing because it was returning a Rhino object.
- It removes an old dependency to `Newtonsoft.json`

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EipGf_yyhclLi5qeS8b_IkIBuQAnJGPmVQlYhup_4jkLew?e=ZsyynE

#### Added:
- `public static Type Type(this IGH_Param param, Caller caller = null)`
- `public override bool DestroyParameter(GH_ParameterSide side, int index)`
- `public void OnGrasshopperUpdates(object sender, IGH_Param param)`
- `public override bool DestroyParameter(GH_ParameterSide side, int index)`

#### Removed:
- `protected override void RegisterInputParams(GH_InputParamManager pManager)`
- `private void SyncParamsFromGH(object sender = null, EventArgs e = null)`

### Additional comments

**N.B. 1** This pr is breaking the (de)serialisation of previous `CreateCustomComponent`s. They do  (de)serialise but the result is incorrect.

**N.B. 2** So, the way the events for creation, destruction and update of the parameters are handled are a little hacky. This is because we are not able create an instance of `Grasshopper.Kernel.GH_ParamServerEventArgs` - its constructor is not accessible.
My first solution was to access the private constructor via reflection and create an instance of it, but I found it more hacky than the current solution. Don't hesitate to chip in, as there may be a better solution.

### DIFF
--- a/Grasshopper_Engine/Convert/FromGoo.cs
+++ b/Grasshopper_Engine/Convert/FromGoo.cs
@@ -22,12 +22,7 @@
 
 using Grasshopper.Kernel.Parameters;
 using Grasshopper.Kernel.Types;
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.Runtime.CompilerServices;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace BH.Engine.Grasshopper
 {
--- a/Grasshopper_Engine/Convert/FromGoo.cs
+++ b/Grasshopper_Engine/Convert/FromGoo.cs
@@ -37,7 +37,7 @@ namespace BH.Engine.Grasshopper
         /**** Public Methods                    ****/
         /*******************************************/
 
-        public static T FromGoo<T>(this IGH_Goo goo, IGH_TypeHint hint=null)
+        public static T FromGoo<T>(this IGH_Goo goo, IGH_TypeHint hint = null)
         {
             if (goo == null)
                 return default(T);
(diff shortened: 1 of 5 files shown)


al-fisher:
@epignatelli Tested this. Seems to now behave as expected - thanks.
In particular great to get the full forced type with a custom data component working again! Really powerful! 👍 

Also - parameter update seems to be behaving as desired too
update names

PR #341, opened 2019-03-26
Issue340 - Improve rendering performance

### LINKED ISSUE #340: Grasshopper_Toolkit: Improve rendering performance of BHoMObjects and IGeometry

epignatelli:

Closes #340 

The rendering methods are now converting back and forth Rhino geometries at every display frame. This causes a big lag when the `Convert` method is computationally expensive.

The conversion is happening in two places:
- The `RenderWires` and `RenderMeshes` methods
- The `get_Value` method of the `BH.Engine.Grasshopper.Objects.*`

This PR is proposing to store the converted rhino geometry and use that one to render and exchange `Value`s.

This is actually modifying the way we pass the values between components, so before merging this, **please, try to break it!**

### Test files

My advice is to try this file using the master and using this branch to check the performance improvement. Then, try to break the code in this branch only.
https://burohappold.sharepoint.com/:f:/s/BHoM/Er0Mwo6YQspNngsbElNVlwABs6T5ejBr_yKnUJxR3ipZ6g?e=dVSLm8
### Additional comments

@adecler FYI

### DIFF
--- a/Grasshopper_Engine/Compute/RenderMeshes.cs
+++ b/Grasshopper_Engine/Compute/RenderMeshes.cs
@@ -49,7 +49,6 @@ namespace BH.Engine.Grasshopper
             {
                 return;
             }
-            Color bhColour = Query.RenderColour(args.Material.Diffuse);
             DisplayMaterial bhMaterial = Query.RenderMaterial(args.Material);
             try
             {
--- a/Grasshopper_Engine/Compute/RenderRhinoMesh.cs
+++ b/Grasshopper_Engine/Compute/RenderRhinoMesh.cs
@@ -1,0 +1,221 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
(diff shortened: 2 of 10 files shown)


IsakNaslundBh:
MASSIVE(!!!) difference in rendering performance, especially for surfaces. Trying with some ~100 PanelPlanar, the performance goes from more or less unusable to close to rhino native performance. Will run on this branch a bit to make sure my scripts works etc before approving, but really liking this!

IsakNaslundBh:
Pushing first fix, not converting to rhino on the data accessor, and leaving that job to the parameters instead.

epignatelli:
OK guys, this should be cleaned up and working now, thanks @IsakNaslundBh for the help.

@IsakNaslundBh if you can do a last a last test to check it is working for you as well, we can merge as soon as it is approved.

I'll check separately the rhino convert

epignatelli:
So, there is a lot of clean up to do I think after this, but things are working now, so I'd rather to fix them in another sprint:

- Since we split the BHoM and the Rhino geometry `Value` is always `IGeometry` from now on. This means `GH_GeometricGoo<object>` will be `GH_GeometricGoo<IGeometry>`
- `CastTo<Q>` method re-tries to convert to the target geometry, whereas we could just get the `m_RhinoGeometry`
- I feel we need more clarity when hovering with our mouse on a parameter. I plan to do something like "BH.oM.Geometry.Line" as "Rhino.Geometry.Line", quite verbose, but better than too cryptic.

PR #344, opened 2019-04-12
Update post build events to centralise assemblies location

### LINKED ISSUE #343: Find a way to load assemblies from the central %AppData%/BHoM folder
Note: this may help https://www.grasshopper3d.com/forum/topics/grasshopper-developer-settings

epignatelli:

Closes #343 

This PR is switching to the [`.ghlink` logic](https://www.grasshopper3d.com/forum/topics/sdk-changes-and-new-features-in-grasshopper-0-9-0050?commentId=2985220%3AComment%3A841407) to load assemblies in Grasshopper.
Achieved though new post build events. 

Note that the post-build events are clearing your %AppData%\Grasshopper\Libraries\BHoM folder and deleting it, in case it exists.

### Test files

Two tests are required: compiling from source and installing through latest BHoM_Installer

### DIFF
--- a/Grasshopper_UI/Grasshopper_UI.csproj
+++ b/Grasshopper_UI/Grasshopper_UI.csproj
@@ -365,9 +365,14 @@
   </Target>
   -->
   <PropertyGroup>
-    <PostBuildEvent>call "$(TargetDir)UI_PostBuild.exe" ..\..\ "C:\Users\$(Username)\AppData\Roaming\Grasshopper\Libraries\BHoM"
+    <PostBuildEvent>call "$(TargetDir)UI_PostBuild.exe" ..\..\ "$(AppData)\BHoM\Assemblies"
+Copy /Y "$(TargetName).dll" "$(AppData)\BHoM\Assemblies\$(TargetName).gha"
 
-Copy /Y "$(TargetName).dll" "C:\Users\$(Username)\AppData\Roaming\Grasshopper\Libraries\BHoM\$(TargetName).gha"</PostBuildEvent>
+IF EXIST $(AppData)\Grasshopper\Libraries\BHoM (
+RD /S /Q $(AppData)\Grasshopper\Libraries\BHoM
+)
+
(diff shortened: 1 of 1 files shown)


alelom:
### Test 1 - compiling from source
Worked. Deleted successfully all the content in `%AppData%\Grasshopper\Libraries\BHoM` and replaced with ghlink file. BHoM loads successfully.

### Test 2 - Installer
I've used the [Master branch for the installer](https://github.com/BuroHappoldEngineering/BHoM_Installer/tree/master).
There are 2 separate issue with the installer here:

1) The installer still puts all the dlls into the `%AppData%\Roaming\Grasshopper\Libraries` instead of a ghlink file pointing to the `%AppData%\Roaming\BHoM` folder.

2) The installer does not put the `BH.UI.Grasshopper.gha` grasshopper plugin file in the `%AppData%\Roaming\BHoM` folder.

awakeman:
On the second point I will need to modify the installer, there is an issue raised to do so and I plan on doing it at the start of next week.

epignatelli:
Thanks for testing this guys! 

What's the status of the installer right now?
If the Installer 2.0 is merged to master, I'd rather wait for it to be fixed before merging the Grasshopper one, or we'll end up with incompatible masters.

awakeman:
It actually won't be incompatible, the installer would just still stick the `.gha` and `.dll`s in the `Grasshopper\Libraries\BHoM` folder just as now. Anyway, I'll let you wait if you wish, I'll be doing the Installer one today, shouldn't take long.

epignatelli:
Then it would actually be incompatible.
You will have doubled the assemblies, the linked from `BHoM\Assemblies` and the ones in `Grasshopper\Libraries\BHoM`. Grasshopper will ask which one to load every time you open it.

Let me know when the installer is done, I'll add the dependency at this PR.

awakeman:
Only if you both run an installer and build it yourself. If you do one or the other then it will either be the new way or the old.

That said I have now made the change for Grasshopper on the Installer but not for Revit and Dynamo.

PR #371, opened 2019-06-06
Grasshopper_Toolkit: 369+370 faster handing of the rendering phase for results

### LINKED ISSUE #369: Grasshopper_Toolkit: Grasshopper slow with lots of data and preview on
When the preview is tuned on when you have a lot of data in gh (with the preview selected to be of for the data-container with the big amounts of data in them) the script is slowed down by a serious amount.

Turning the global preview off makes the script run quick again.

This is particularly problematic when working with Results where you easily could end up with millions loaded at the same time.

Turning "IsPreviewable" for the parameter to false solves the problem.

One solution would be to add another Goo and Parameter for the `IResult` class with the setting as mentioned above. Quick tests with native GH components seem to indicate that the problem is in the actual system, and nothing we can fix locally.

See testfiles:

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/Grasshopper_Toolkit/Grasshopper_Toolkit-Issue369-SlowPreview?csf=1&e=thFXdh

### LINKED ISSUE #370: Grasshopper_Toolkit: Seting output fails if any geometry operations fails
Currently the SetGeometry() is called by the Value property for the Goo.

I think this is the right place to do it, but it needs to sit in a try-catch to ensure that the value can be set correctly even if geometry extraction/convertion fails. 

Suggesting a try-catch around the SetGeometry() here:

image

epignatelli:

Closes #369 
As described #369 even if objects to not have a geometry, but its property `IsPreviewCapable` == `true` Grasshopper till includes it when redrawing (something, what exactly, we have no idea) viewports.
When a parameter contains 1e6 objects, that gets slow.
With this pr we are explicitly setting the `IsPreviewCapable` property to false, in case the parameters contains `IResult`. Specifically, we are checking only the first object.
We can extend the number of objects to exclude from preview in the `IsPreviewExcluded` method from the Engine.

Closes #370
Rendering geometry should not stop the workflow if it fails. https://github.com/BHoM/Grasshopper_Toolkit/commit/c50b3b69c998f7feed778914fffe24cb3355fa5e is adding a `try` `catch` when geometry is set for both `GH_IObject` and `GH_BHoMObject` types.
Some experiments about performances [here](https://stackoverflow.com/questions/1308432/do-try-catch-blocks-hurt-performance-when-exceptions-are-not-thrown).
I didn't log an error message when an exception is caught to not pollute the component.
If the method ran correctly, but rendering failed, the user might be mislead.

### Test files

To test, you can try before and after this pr to add a component on the canvas after opening the following file. It will take a significantly different time.
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/Grasshopper_Toolkit/Issue369-IsPreviewExcluded/Issue369-IsPreviewExcluded.gh?csf=1&e=eQ6ciV


Grasshopper_UI/2.1/Components/Parameters/Param_IObject.cs
```diff
@@ -38,6 +39,14 @@ public class Param_IObject : BHoMParam<Engine.Grasshopper.Objects.GH_IObject>
 
         public override string TypeName { get; } = "IObject";
 
+        public override bool IsPreviewCapable
+        {
+            get
+            {
+                return !m_data.get_FirstItem(true)?.Value.IsPreviewExcluded() == true;
```
adecler: I wonder if we shouldn't simply disable display if the number of items is too high since this is the main reason for doing this in the first place. We could then throw a single warning stating that display has been disable.  To be completely user friendly, we should then give the option to the user (as a tick box in the menu) to keep display on regardless of internal checks.

A few minor additional comments on that line:
- Not need for the '== true'. Comparing a boolean to true is equivalent to returning the boolean itself
- The double negation makes this harder to read. Why not just `m_data.get_FirstItem(true)?.Value.IsPreviewCapable()`

Grasshopper_Engine/Query/IsPreviewExcluded.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace BH.Engine.Grasshopper
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static bool IsPreviewExcluded(this object value)
```
adecler: I have a few issues with this:
- If such a method should exist, it should be in the BHoM_UI, not in GH.
- The listing of excluded type as it is done here is very bespoke to the problem you are currently facing and doesn't scale well at all. First, you are taking the risk of adding dependencies to dlls the UI should not be aware of. Secondly, this completely moves a part of the control of objects from the oM and into a section of the code far less accessible. See my comment below for an alternative option. If something needs to be handle on type by type bases, then it needs to appear in the oM as attribute, not hidden in the UI.

Grasshopper_UI/2.1/Components/Parameters/Param_IObject.cs
epignatelli: I like the solution based on on data count.

For the code itself, `== true` is necessary since `m_data.get_FirstItem(true)?.Value.IsPreviewCapable()` returns a reference type (the null propagation operator).

Grasshopper_Engine/Query/IsPreviewExcluded.cs
epignatelli: I removed the method altogether, although the idea was to use the same pattern we use for `Query.Geometry`. it was here since the need was not yet arisen.

Grasshopper_UI/2.1/Components/Parameters/Param_IObject.cs
epignatelli: Pushed it - I limited the behaviour to the `GH_IObject` parameter, It is unlikely that it happens for all the others, so very localised.

Grasshopper_UI/2.1/Components/Parameters/Param_IObject.cs
```diff
@@ -38,6 +39,23 @@ public class Param_IObject : BHoMParam<Engine.Grasshopper.Objects.GH_IObject>
...
         public override string TypeName { get; } = "IObject";
 
+        public override bool IsPreviewCapable
+        {
+            get
+            {
+                if (m_ForcePreview != null)
+                    return m_ForcePreview == true;
+
+                if (VolatileDataCount > 100000 && m_ForcePreview == null)
```
adecler: Always store those hard numbers into a variable and never use them directly. I know you are only using it in one place in this case so it might not feel as relevant to do that. But it is important to keep that as a good practice everywhere.

On a side note, it feels like 100k is too high, I would have happily settled for 10k. You can always set it to 10k and let the user change that in the menu of the component (another benefit of having it as a variable ;-p)

Grasshopper_UI/2.1/Templates/BHoMParam.cs
```diff
@@ -103,6 +103,16 @@ public virtual void OnSourceCodeClick(object sender = null, object e = null)
...
         }
 
+        /*******************************************/
+
+        public virtual void OnPreviewCapableClick(object sender = null, object e = null)
+        {
+            if (m_ForcePreview == null)
+                m_ForcePreview = true;
+            else
+                m_ForcePreview ^= true;
```
adecler: That is not a very common operator. Remember we have to stay as readable as possible. I also don't get why you are not using a simple boolean for `m_ForcePreview` whith default value of false. This makes the whole thing more complicated that it needs to be.

Grasshopper_UI/2.1/Templates/BHoMParam.cs
```diff
@@ -151,10 +161,17 @@ public override void AppendAdditionalMenuItems(ToolStripDropDown menu)
...
 
-        /*******************************************/
-    }
 
-    /*******************************************/
+        /***************************************************/
+        /**** Private Fields                            ****/
+        /***************************************************/
+
+        protected bool? m_ForcePreview = null;
```
adecler: Same thing here.

adecler:
> Getting below the threshold at 99k forces, you start feeling a big slowdown again.

@IsakNaslundBh , then it means 99k is too high. You would have the same issue with any other object. `IResult` is just the only one that we have at the moment with a regular high count but the problem at hand has nothing to do with the `IResult` class itself. Sure,  `IResult` doesn't have a geometry but would you be happy with the delay created by drawing 99k geometrical objects ? The answer is probably 'no' in most cases. For the ones when you do, you have the tick box to force display.

epignatelli:
Ok, thanks guys for the feedback.
@adecler as per our last conversation this morning:
- The user has control over the threshold to display
- If the number of items is still high, you can "preview off" the parameter and it will set `IsPreviewCapable` to false.

I also handled (de)serialisation - checked we have no issues with old parameters, that should be ok.

@IsakNaslundBh does this solution sound good?
Let me know if anything goes wrong in your test.

epignatelli:
@IsakNaslundBh @adecler if you're happy with this, can you approve?

epignatelli:
Thanks for the green stamp @IsakNaslundBh, waiting for the BHoM call to merge this.

adecler:
I am not a big fan of the variable that takes both roles of storing the max number of items AND forcing the display. Fells a bit hacky and it makes the menu look horrible. I would rather have two variables in this case.

epignatelli:
I just requested @alelom for a review, to unblock this. The only test I asked for is a compilation test, as this as been reviewed enough.

Should there be any issue about this merge, do reach out to me!

PR #405, opened 2019-09-06
Solves #396

### LINKED ISSUE #396: Grasshopper_Toolkit: Rhino crash with custom component with old types
#### Description:

If you have a `CreatCustom` component on the canvas targeting a type that is then changed (name/namespace), deleteing this component causes a full Rhino crash.

After investigation with @epignatelli the root of the problem seem to be a missing null check for types making the `CompileInputGetters` in the BHoM_UI crash.

Updating method in `Type` method in Grasshopper_Engine to

```

     public static Type Type(this IGH_Param param, Caller caller = null)
        {
            Type type;

            if (param == null)
                type = typeof(object);

            if (param is Param_ScriptVariable)
                type = Type(((Param_ScriptVariable)param).TypeHint, param.Access);

            else if (caller != null)
                type = (caller.InputParams.Find(x => x.Name == param.NickName))?.DataType;

            else
                type = param.Type;

            if (type == null)
                return typeof(object);
            else
                return type;
        }
```
stops the crash, but not sure if it has further implications.

#### How to replicate:

1. Create a BHoMObject using CustomObject
2. Change the object name/namespace
3. try delete/update the component

#### Expected behaviour:

For rhino not to crash.

epignatelli:

Closes #396 

Rhino crashes altogether if you open a script with a  `CustomObjecy` component that serialised a `Type` that doesn't exist anymore in the BHoM.

### Test files

@IsakNaslundBh would you mind using the file that triggered the issue in the first place?

### Changelog

- Make sure that `Engine.Grasshopper.Query.Type(this IGH_Param param, Caller caller = null)` never returns `null`.

### Additional comments

This is quite critical, since it is crashing the whole Rhino.

### DIFF
--- a/Grasshopper_Engine/Query/Type.cs
+++ b/Grasshopper_Engine/Query/Type.cs
@@ -38,17 +38,18 @@ namespace BH.Engine.Grasshopper
 
         public static Type Type(this IGH_Param param, Caller caller = null)
         {
+            Type type;
             if (param == null)
-                return typeof(object);
+                type = typeof(object);
 
             if (param is Param_ScriptVariable)
-                return Type(((Param_ScriptVariable)param).TypeHint, param.Access);
-
+                type = Type(((Param_ScriptVariable)param).TypeHint, param.Access);
             else if (caller != null)
-                return (caller.InputParams.Find(x => x.Name == param.NickName))?.DataType;
-
+                type = caller.InputParams.Find(x => x.Name == param.NickName)?.DataType;
             else
-                return param.Type;
+                type = param.Type;
+
+            return type == null ? typeof(object) : type;
         }
 
         /***************************************************/


epignatelli:
From the Sprint Closure call @IsakNaslundBh to have a go with the script creating the issue

epignatelli:
@IsakNaslundBh shall we have a look at this together, tomorrow morning?

Grasshopper_Engine/Query/Type.cs
```diff
@@ -38,17 +38,18 @@ public static partial class Query
 
         public static Type Type(this IGH_Param param, Caller caller = null)
         {
+            Type type;
             if (param == null)
-                return typeof(object);
+                type = typeof(object);
```
adecler: This one needs to stay a `return` as it doesn't belong to the same if-else chain as the rest.

PR #426, opened 2019-10-23
Update to the new version of the CreateObject component

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_UI/pull/147

   
See https://github.com/BHoM/BHoM_UI/pull/147 for more details


epignatelli:
Quoting the other comment https://github.com/BHoM/BHoM_UI/pull/147#issuecomment-545348650

Grasshopper_UI/Components/oM/CreateObject2.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+
+        /*******************************************/
+        /**** Override Methods                  ****/
+        /*******************************************/
+
+        public override bool CanInsertParameter(GH_ParameterSide side, int index)
+        {
+            return side == GH_ParameterSide.Input;
```
epignatelli: If you select a `Create` method from the menu, the component still leaves the possibility to remove and add parameters. If you remove one it crashes. Maybe we can prevent it from happening?
```suggestion
            return side == GH_ParameterSide.Input && !(Caller?.SelectedItem is MethodInfo);
```

Grasshopper_UI/Components/oM/CreateObject2.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        /*******************************************/
+
+        public override bool CanRemoveParameter(GH_ParameterSide side, int index)
+        {
+            return side == GH_ParameterSide.Input;
```
epignatelli:
```suggestion
            return side == GH_ParameterSide.Input && !(Caller?.SelectedItem is MethodInfo);
```

Grasshopper_UI/Components/oM/CreateObject2.cs
adecler: Thanks. Yes, that was on the todo list for today but good that you are picking this up.
Now I can just copy-paste 😄

adecler:
This is now ready for review

Grasshopper_UI/Components/oM/CreateObject.cs
```diff
@@ -26,10 +26,11 @@
 using BH.UI.Grasshopper.Templates;
 using BH.UI.Templates;
 using BH.UI.Components;
+using System.Reflection;
 
 namespace BH.UI.Grasshopper.Components
 {
-    public class CreateObjectComponent : CallerComponent
+    public class CreateObjectComponent : CallerComponent, IGH_VariableParameterComponent
```
epignatelli: I think there is no need to inherit from `IGH_VariableParameterComponent`.
`CallerComponent` already does.

Grasshopper_UI/Components/oM/CreateObject.cs
```diff
@@ -38,6 +39,32 @@ public class CreateObjectComponent : CallerComponent
         public override Caller Caller { get; } = new CreateObjectCaller();
 
 
+        /*******************************************/
+        /**** Override Methods                  ****/
+        /*******************************************/
+
+        public override bool CanRemoveParameter(GH_ParameterSide side, int index)
```
epignatelli: Very much related to https://github.com/BHoM/BHoM_UI/pull/147/files#r341051725 for the missing `CanCreateParameter`.

It may be annoying from a UX perspective to erroneously remove an input, realising it after a series of operation, and could not CTRL + Z anymore.

epignatelli:
Approving in the light of https://github.com/BHoM/BHoM_UI/pull/147#issuecomment-548343169

#### Please, do not merge this.

Grasshopper_UI/Components/oM/CreateObject.cs
adecler: Hmmm weird, I remember having to add this after having problems with the inputs not updating correctly. But you're right it shouldn't need to be there. It doesn't do any damage either way so I guess its fine :-p

Grasshopper_UI/Components/oM/CreateObject.cs
adecler: Yes, we discussed the idea of adding a 'reset' button in the context menu but I like your idea of the full list in the context menu better. Especially when considering the fact that it allows us to not have the base properties by default (except `Name` I guess)

adecler:
@epignatelli , not a big deal as we will discus soon anyway but not actually approved :-p

PR #442, opened 2019-11-15
Add support for lists as input default values

### LINKED ISSUE #441: Engine Create methods don't work on lists
The new create components do not appear to handle list objects quite the way as is expected.

I tried to create a new panel object in Grasshopper, and when plugging in my edges I got the error:
 > 1. This component failed to run properly. Inputs cannot be collected properly.
Cannot convert type 'System.Collections.Generic.List<BH.oM.Environment.Elements.Edge>' to 'BH.oM.Environment.Elements.Edge' 

I got the same on an Environment.Opening object (with edges again), and @IsakNaslundBh was getting the same on some of his structural objects.

It appears that objects that don't have collection properties work fine (e.g. Bar appears to be working ok), it's just something up with the collection properties.

#### How to replicate:
Try to create any object which has a collection property, e.g.:
 - Environment.Elements.Panel
 - Environment.Elements.Opening
 - Environment.Elements.Space
 - Structure.Elements.Panel

An error occurs in trying to cast the list input to a single object.

#### Expected behaviour:
That the error does not occur and the object can be created successfully.

#### Test file(s):
Being uploaded, will edit issue with link when it uploads.
[Test script](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_UI/BHoM_UI-%23165-CreateMethods.gh?csf=1&e=kw0o5t)

I have assigned @adecler initially because I'm unsure who will be best to tackle it, but @adecler please reassign as necessary :smile:

adecler:

Closes #441 

### Test files
- See #441
- The key thing here is to make sure this PR doesn't break anything. 

### Additional comments
I have only added support for lists and lists of lists. We could extend to more IEnumerables (other than dictionary) but the list is the one that is officially supported so I focused on that one. Given that this is a big change already and has a potential risk of breaking something I thought I would keep it as simple as possible.

I am aware that HashSet is also failing but we don't have a GH_HashSet class at the moment (as we have for GH_Dictionary). So I think this is better to fix the current bug of this issue and solve the HashSet in another PR (Especially given the current number of PRs in this repo!)

### DIFF
--- a/Grasshopper_Engine/Convert/ToGoo.cs
+++ b/Grasshopper_Engine/Convert/ToGoo.cs
@@ -23,6 +23,7 @@
 using BH.Engine.Grasshopper.Objects;
 using BH.oM.Base;
 using BH.oM.Geometry;
+using Grasshopper.Kernel.Data;
 using Grasshopper.Kernel.Types;
 using System;
 using System.Collections;
@@ -47,6 +48,26 @@ namespace BH.Engine.Grasshopper
                 return ToGoo(obj as dynamic);
         }
 
+        /*************************************/
+
+        public static List<IGH_Goo> IToGoo<T>(this List<T> list)
+        {
+            return list.Select(x => x.IToGoo()).ToList();
+        }
+
+        /*************************************/
+
+        public static GH_Structure<IGH_Goo> IToGoo<T>(this List<List<T>> tree)
+        {
+            GH_Structure<IGH_Goo> structure = new GH_Structure<IGH_Goo>();
+
+            for (int i = 0; i < tree.Count; i++)
+                structure.AppendRange(tree[i].IToGoo(), new GH_Path(i));
+
+            return structure;
+        }
+
+
         /*******************************************/
         /**** Public Methods                    ****/
         /*******************************************/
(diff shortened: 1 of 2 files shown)


Grasshopper_Engine/Convert/ToGoo.cs
```diff
@@ -47,6 +48,26 @@ public static IGH_Goo IToGoo(this object obj)
                 return ToGoo(obj as dynamic);
         }
 
+        /*************************************/
+
+        public static List<IGH_Goo> IToGoo<T>(this List<T> list)
```
epignatelli: Shouldn't this be a non-interface method?

Grasshopper_Engine/Convert/ToGoo.cs
```diff
@@ -47,6 +48,26 @@ public static IGH_Goo IToGoo(this object obj)
...
+        {
+            return list.Select(x => x.IToGoo()).ToList();
+        }
+
+        /*************************************/
+
+        public static GH_Structure<IGH_Goo> IToGoo<T>(this List<List<T>> tree)
```
epignatelli: Ditto

Grasshopper_Engine/Convert/ToGoo.cs
adecler: Well, the problem is I need this to be a `I` method so it is a common method name for objects and lists. This is a bit of a stretch but I see it as the I is inside the list, not outside.

adecler:
Thanks @FraserGreenroyd , I have merged this because there are far too many PRs on this repo clogging up the whole thing (and more to come)

PR #449, opened 2020-01-09
Fix composition loop for DataAccessor

### LINKED ISSUE #448: DataAccessor implementations create a closed encapsulation loop
In the Grasshopper implementation of the `DataAccessor`, the `DataAccessor_GH` owns a `GH_Component`, and the `CallerComponent` owns a `DataAccessor_GH`. 

This closed loop makes serialisation of a `Caller`object to fail (`Caller.ToJson()`), since it iterates infinitely between component and `DataAccessor`.

It may required modifications in the implemented UIs as well, so it's may be not circumscribed to the BHoM_UI

adecler:

Closes #448 

I have removed the dependency to the Caller by only passing the Component Inputs to the DataAccessor as it is the only thing needed anyway.

### DIFF
--- a/Grasshopper_UI/Templates/CallerComponent.cs
+++ b/Grasshopper_UI/Templates/CallerComponent.cs
@@ -76,7 +76,7 @@ namespace BH.UI.Grasshopper.Templates
             Category = "BHoM";
             SubCategory = Caller.Category;
 
-            Accessor = new DataAccessor_GH(this);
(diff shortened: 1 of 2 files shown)


adecler:
Ha, I forgot to do that change before pull requesting.
I don't think that should be left on the master in that state. 
Since you merged already, I will do another PR.

PR #453, opened 2020-01-10
Allow to remove outputs for the Explode component and show BHoM menu on wire drag-and-released event

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_UI/pull/186

   
https://github.com/BHoM/BHoM_UI/issues/183
https://github.com/BHoM/BHoM_UI/issues/196

See comments in https://github.com/BHoM/BHoM_UI/issues/186

### Changelog
- Allow to remove outputs for the Explode component
- Show BHoM menu on wire drag-and-released event

### Additional comments
Purely for discussion purposes at the moment.


epignatelli:
Before expressing an opinion, can I ask what is the actual benefit/behaviour we are aiming at with enabling this?

adecler:
@epignatelli , it's pretty simple: most of the time the Explode component is used to exact only a small amount of properties from an object. It can however sometimes be quite large (take for example exploding MEP objects). A few times, I have resorted to use the `GetProperty` component for that very reason. But the `GetProperty` component is not as nice due to the need to provide a string for the property and becomes messy once you need to get multiple properties.

Allowing to remove outputs from the Explode component therefore seems like the obvious step to take. It also mirrors nicely the new functionality of the `Create` component.

epignatelli:
Thanks @adecler! If that's the case what I would suggest is to use the same solution I proposed for the `CreateObject` with object initializers.

Instead of adding and removing wiht the `+` and `-` signs, I think for the user it would be very useful to have the `Add/Remove Outputs` option in the context menu. The reason is that you can add back if you want, or have a partial list of outputs whilst reading the list of all possible outputs.

What do you think?

al-fisher:
Similar to the Create - I still advocate for (_only_) the minus symbols directly on the component to remove outputs you are not interested in. But agree addition of outputs needs context menu to help the user know what can actually be added. Is that what you meant @epignatelli ?

The explode is a really heavily used component - I frequently use the same way @adecler describes above. Also for new users (or when inspecting new objects) the explode is a great way of just exploring what makes up the object! Used a lot in teaching/explaining. One of the first components.
Thus would certainly recommend smoothing the user experience for both new _and_ frequent users.
(I am lazy! 🙈 I don’t want to right click for menu to remove an output! 🐒 )

Thus perhaps both the minus on the component (for quick access) and then also context menu for more control? This is in fact what we have for the Create right?

epignatelli:
Yes! That's what I am proposing!
And I completely agree on the propedeutic use of the `Explode` to inform, rather than do!

I have nothing against the `-` sign, but trying to project myself in the moment of using it, I think I will not use it. I understand many will, though, so happy to keep it!

My point is that, if I am lazy (I am even lazier, maybe 😄) I would not even bother clicking minuses. If there are a lot of outputs, clicking all those minuses is going to be very tedious (as it is going to be wiht the context menu as well).
If we want to ease that up, I'd look for something that makes me click on what I want to keep, rather than what I want to remove? Or maybe both, depending on the proportions of the the sets.

All of this is not solved with the context menu anyway, but that gives control and knowledge about the object, I think.

al-fisher:
> My point is that, if I am lazy (I am even lazier, maybe 😄) I would not even bother clicking minuses. 

😆 

Yes - a select/deselect all check box at top of context menu is an option?

rwemay:
Great discussion. I personally don’t see the harm with minus on explode, as long as ‘update’ reinstates all. I think I’d agree with not complicating further with plus etc. 

I probably would use this, to keep scripts readable - assuming copy paste of a reduced explode didnt automatically reset to all properties (way too much minus clicking).

The tick box context menu I think, would be a great addition. GetProperty is great but has the disadvantage of needing to know property names, and always showing the prop name in the script.

epignatelli:
> > My point is that, if I am lazy (I am even lazier, maybe smile) I would not even bother clicking minuses.
> 
> laughing
> 
> Yes - a select/deselect all check box at top of context menu is an option?

Yes! I was thinking at click-and-drag to select/deselect multiple entries, but I agree with a checkbox Select/Unselect all!

adecler:
I have now mirrored the input/output selection menu from the CreateObject component:

image

Note that The Explode component updates its outputs only when the menu closes to avoid computation every time an item is (un)ticked. We could do the same for the CreateObject component.

Also note that I have made sure that the menu stays open after toggling an item. I though there was an issue raised on that but cannot seem to be able to find it.

rwemay:
Nice! @jdarcyBHE  see here also FYI, apologies @epignatelli , I had not picked up on the create component add/remove menu item in lieu of +/- buttons. Great stuff, thanks guys!!

adecler:
Now ready for review and merging

adecler:
Thanks @IsakNaslundBh , 
- number 1 should be pretty straightforward to fix. I'll have a look now.
- number 2 is a trade off I was aware of but I am more than happy to go in your direction (one line of code). The problem is All non-BHoM inputs/outputs will not trigger the menu even if they are on BHoM components. I kept the menu to reassure users that something was happening but I guess an empty menu is not much better than no menu at all.

IsakNaslundBh:
> number 2 is a trade off I was aware of but I am more than happy to go in your direction (one line of code). The problem is All non-BHoM inputs/outputs will not trigger the menu even if they are on BHoM components. I kept the menu to reassure users that something was happening but I guess an empty menu is not much better than no menu at all.

Ah, yeah. Get your point there. Personally I would prefer to not have to hit escape every time you grab the wrong output by mistake to close the menu on a non-BHoM component, but see your point about consistency.

adecler:
I have pushed the two fixes as separate commit. 
As you can see it is easy to make the change either way for 2 so happy to go with the majority

Grasshopper_UI/Global/GlobalSearch.cs
```diff
@@ -75,18 +137,74 @@ private static void GlobalSearch_ItemSelected(object sender, oM.UI.ComponentRequ
...
+                                    Type sourceType = GetSourceType(x);
+                                    return sourceType != null && GetSourceType(x).IsAssignableFrom(m_LastWire.SourceType);
+                                }); 
+                            if (param != null)
+                                param.AddSource(m_LastWire.Source);
+                        }
+                    }
+
+                    canvas.Invalidate();
+                    component.ExpireSolution(true);
```
epignatelli: This will break if `component` is `null`., it happens with `Enum`s. You can try creating a `Push` component and dragging from `PushType` to reproduce.

adecler: Nice catch!
I have updated the code to cover both `GH_Component` and `IGH-param` components and making sure they all get linked properly

Grasshopper_UI/Global/GlobalSearch.cs
```diff
@@ -53,11 +56,70 @@ public static void Activate()
...
+
+        private static void Canvas_MouseUp(object sender, System.Windows.Forms.MouseEventArgs e)
+        {
+            GH_Canvas canvas = sender as GH_Canvas;
+            if (canvas == null)
+                return;
+
+            if (m_LastWire == null)
+                return;
+
```
epignatelli:
```suggestion
            if (e.Button != System.Windows.Forms.MouseButtons.Left)
                return;

```

Grasshopper_UI/Global/GlobalSearch.cs
```diff
@@ -53,11 +56,70 @@ public static void Activate()
...
+            FieldInfo targetField = typeof(GH_WireInteraction).GetField("m_target", BindingFlags.NonPublic | BindingFlags.Instance);
+            if (targetField != null && targetField.GetValue(m_LastWire.Wire) != null)
+                return;
+
+            GlobalSearch.Open(canvas.FindForm(), m_LastWire.SourceType, m_LastWire.IsInput);
         }
```
epignatelli:
```suggestion

        /*******************************************/
        private static void Canvas_KeyDown(object sender, System.Windows.Forms.KeyEventArgs e)
        {
            GH_Canvas canvas = sender as GH_Canvas;
            if (canvas == null)
                return;

            if (e.KeyData == System.Windows.Forms.Keys.Escape)
                m_LastWire = null;

            return;
        }

```

PR #476, opened 2020-03-05
Fix incorrect wire creation on component created from global menu

### LINKED ISSUE #475: Undesired wire created after some call to the global menu
#### Description:

- when creating a component with ctrl+shift+b, sometimes the input wires are being randomly(?) connected to the existing components

The issue seems to happen quite randomly and we did not find a way to isolate/reproduce it. This does not help (🙈), but it still would be great to get this one resolved in the current milestone as sometimes it gets really annoying and confusing for the end user.

@IsakNaslundBh please feel free to change/supplement this message if I missed something important.

adecler:

Closes #475

The bug was coming from line 120 in the `GlobalSearch` file. When a wire was connected to the input of another component, the global menu was not opened (as expected) but the wire was not cleared. So if you use the global menu directly after that, it would still use the old wire.

### Additional comments
Note that I have also added a try-catch in the `GlobalSearch_ItemSelected` method to make sure that the wire will always be cleared no matter what. This method is called when an item is selected but also when the menu loses focus so there shouldn't be any possible situation where a wire is created but not deleted now.

Finally, I have added a few debug messages that should help in case further debugging is needed. I suggest we keep them there until we are 100% that we have no more bug related to menu selection. This could also potentially help with identifying the other two bugs related to the global menu found recently.

### DIFF
--- a/Grasshopper_UI/Global/GlobalSearch.cs
+++ b/Grasshopper_UI/Global/GlobalSearch.cs
@@ -37,6 +37,7 @@ using GH = Grasshopper;
 using Grasshopper.GUI.Canvas.Interaction;
 using Grasshopper.Kernel;
 using BH.Engine.Reflection;
+using System.Diagnostics;
 
 namespace BH.UI.Grasshopper.Global
 {
@@ -101,6 +102,8 @@ namespace BH.UI.Grasshopper.Global
                     SourceType = sourceType,
                     IsInput = isInput
                 };
+
+                Debug.WriteLine(DateTime.Now.ToString("HH:mm:ss") + "- Created wire info.");
             }            
         }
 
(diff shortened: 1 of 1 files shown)


adecler:
@pawelbaran , @IsakNaslundBh , I have simplified the corresponding issue the incorrect creation of wires and move all other bugs raised there to separate issues.

pawelbaran:
@adecler I will run on this branch for some time to make sure the issue does not occur any more. @IsakNaslundBh @kThorsager maybe you could also switch?

IsakNaslundBh:
@pawelbaran yep, already on the branch, was planning to run a little while to make sure it is gone, but have not had any problems since I switched over.

adecler:
@pawelbaran , @IsakNaslundBh , does this PR seem to fix the wire issue for you ? If not, please let me know as soon as possible as I will need time to work on this before Friday.

kThorsager:
I have not noticed any issues after switching and was experiencing it a lot beforehand

PR #490, opened 2020-03-18
Fix message notes covering all other messages

### LINKED ISSUE #480: Note covers all other messages
#### Description:
When Reflection records a Note, it appears on top of all other messages incl. errors, also overriding the balloon colour:

image

Shouldn't the hierarchy be:
- Error
- Warning
- Note

IsakNaslundBh:

Closes #480 

Think this resolves the issue, in a less than elegant way.

So problem seem to stem from the `RuntimeMessageLevel` method on the base `GH_Component` and or `GH_ActiveObject`. This has the current behaviour:

1. `Error` added before any `Remark` -> `Error`
1. `Error` added after `Remark` -> `Remark`
1. `Error` added before `Warning` -> `Error`
1. `Error` added after `Warning` -> `Error`
1. `Warning` added before `Remark` -> `Remark`
1. `Warning` added after `Remark` -> `Remark`

Basically, error trumps everything, if it is higher up in the list, if not, Remark trumps the rest. Problem seem to be the enum itself having a quite weird value indexing, which then fail when comparing:

```
    //
    // Summary:
    //     Indicates the severity of a runtime message.
    public enum GH_RuntimeMessageLevel : byte
    {
        //
        // Summary:
        //     No warnings or errors were recorded during solver-time processes
        Blank = 0,
        //
        // Summary:
        //     One or more warnings (but no errors) were recorded during solver-time processes
        Warning = 10,
        //
        // Summary:
        //     One or more errors (and possibly any number of warnings) were recorded during
        //     solver-time processes.
        Error = 20,
        //
        // Summary:
        //     One of more messages (but no warnings and no errors) were recorded during solver-time
        //     processes.
        Remark = 255
    }
```
(Also note that the description is wrong, as we have experienced).

Now, on the GH_ActiveObject, the `RuntimeMessageLevel` is virtual, so hoped to be able to override it, but nono, that can not be done, cause the GH_Component has it marked as `sealed`.

So some solutions I could think of was

1. The one implemented, force add notes as blanks instead of remarks, if any warnings/errors are found. If noting is found, add as remark, to make sure they show up on the menu.
2. override the GH_ActiveObejct ourselfs, instead of the GH_Component, but that mean adding in a LOT of boilerplate code currently handled by the GH_Component.
3. Override the GH_CompoentAttribute class to handle the message level from there. This would be quite deep in everything though, as we would need to override the general rendering methods etc.

Hence, even though I am not a massive fan of the proposed solution, I think it might be the best one I can think of for now...

### DIFF
--- a/Grasshopper_UI/Others/Logging.cs
+++ b/Grasshopper_UI/Others/Logging.cs
@@ -40,15 +40,22 @@ namespace BH.UI.Grasshopper.Others
         {
             if (events.Count > 0)
             {
-                foreach (Event e in events)
-                {
-                    if (e.Type == EventType.Error)
-                        component.AddRuntimeMessage(GH_RuntimeMessageLevel.Error, e.Message);
-                    else if (e.Type == EventType.Warning)
-                        component.AddRuntimeMessage(GH_RuntimeMessageLevel.Warning, e.Message);
-                    else if (e.Type == EventType.Note)
-                        component.AddRuntimeMessage(GH_RuntimeMessageLevel.Remark, e.Message);
-                }
+                var errors = events.Where(x => x.Type == EventType.Error);
+                var warnings = events.Where(x => x.Type == EventType.Warning);
+                var notes = events.Where(x => x.Type == EventType.Note);
+
+                foreach (Event e in errors)
+                    component.AddRuntimeMessage(GH_RuntimeMessageLevel.Error, e.Message);
+
+                foreach (Event e in warnings)
+                    component.AddRuntimeMessage(GH_RuntimeMessageLevel.Warning, e.Message);
+
+                GH_RuntimeMessageLevel noteLevel = GH_RuntimeMessageLevel.Remark;
+                if (errors.Count() > 0 || warnings.Count() > 0)
+                    noteLevel = GH_RuntimeMessageLevel.Blank;
+
+                foreach (Event e in notes)
+                    component.AddRuntimeMessage(noteLevel, e.Message);
             }
         }
 
--- a/Grasshopper_UI/Others/Logging.cs
+++ b/Grasshopper_UI/Others/Logging.cs
@@ -24,6 +24,7 @@ using BH.oM.Reflection.Debugging;
 using Grasshopper.Kernel;
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


alelom:
Given the constraints, I think you did an excellent job and `1` is the best solution currently achievable. If `1` works, then I think we should not lose our time behind something that is clearly just a bug of Grasshopper. Testing now.

alelom:
This is the code that I used for the test components:
<details> 

```cs
        public static void ErrorTest1()
        {
            BH.Engine.Reflection.Compute.RecordError("Error");
            BH.Engine.Reflection.Compute.RecordWarning("Warning");
            BH.Engine.Reflection.Compute.RecordNote("Note");
        }

        public static void ErrorTest2()
        {
            BH.Engine.Reflection.Compute.RecordWarning("Warning");
            BH.Engine.Reflection.Compute.RecordError("Error");
            BH.Engine.Reflection.Compute.RecordNote("Note");
        }

        public static void ErrorTest3()
        {
            BH.Engine.Reflection.Compute.RecordNote("Note");
            BH.Engine.Reflection.Compute.RecordError("Error");
            BH.Engine.Reflection.Compute.RecordWarning("Warning");
        }

        public static void ErrorTest4()
        {
            BH.Engine.Reflection.Compute.RecordNote("Note");
            BH.Engine.Reflection.Compute.RecordWarning("Warning");
            BH.Engine.Reflection.Compute.RecordError("Error");
        }

        public static void ErrorTest5()
        {
            BH.Engine.Reflection.Compute.RecordWarning("Warning");
            BH.Engine.Reflection.Compute.RecordNote("Note");
            BH.Engine.Reflection.Compute.RecordError("Error");
        }

        public static void WarningTest1()
        {
            BH.Engine.Reflection.Compute.RecordWarning("Warning");
            BH.Engine.Reflection.Compute.RecordNote("Note");
        }

        public static void WarningTest2()
        {
            BH.Engine.Reflection.Compute.RecordNote("Note");
            BH.Engine.Reflection.Compute.RecordWarning("Warning");
        }

        public static void NoteTest()
        {
            BH.Engine.Reflection.Compute.RecordNote("Note");
        }
```

</details>

al-fisher:
Bravo @alelom 😸

adecler:
If I may, this looks awfully complicated to me. 
Why not just skip the notes if there are warnings or errors to display ?

IsakNaslundBh:
> If I may, this looks awfully complicated to me.

You may, and I know it is awfully complicated, but felt the simplest of the various options I could think of. The message class is private, and the field is private, hence all the bloody reflection. But happy to chat through.

> Why not just skip the notes if there are warnings or errors to display ?
Do not think this is a good idea. If the method creator has made the effort to put a note down on something, I do;t think that should just disappear because an warning is raised elsewhere. Especially for things like adapters you might have a large number of methods called, which all could give warnings/notes/errors.

IsakNaslundBh:
@alelom Restructuring code after a chat with @adecler . Also added comments clarifying why this was done.

Would you mind re-reviewing, everything should work exactly as before.

PR #492, opened 2020-03-21
Intellisense for non-BHoM components

### LINKED ISSUE #491: Provide intellisense for all Grasshopper component
#### Description:
This means:
- adding all components in the Ctrl-Shift-B menu
- Wire menu support for all components

This includes components found in installed plugins

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_UI/pull/229

   
Closes #491

### Additional comments
- @emidio-piermarini , this still needs some work and a lost of testing but here it is so you can start playing with it :smile:
- This can only be merged in the next quarter (3.2)

### DIFF
--- a/Grasshopper_UI/Global/GlobalSearch.cs
+++ b/Grasshopper_UI/Global/GlobalSearch.cs
@@ -38,6 +38,8 @@ using Grasshopper.GUI.Canvas.Interaction;
 using Grasshopper.Kernel;
 using BH.Engine.Reflection;
 using System.Diagnostics;
+using Grasshopper.GUI.Ribbon;
+using BH.oM.UI;
 
 namespace BH.UI.Grasshopper.Global
 {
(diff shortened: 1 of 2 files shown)


rwemay:
Wow. Yes. This will help with quite a few things, including knowing which package (non GH native) you're using a component from. Nice one.

adecler:
- Added support for GH_Point (for some reason they are the only non-generic Goo types in GH)
- giving priority of components of the same package (use the concept of tags to increase search weights)
- improve the way the content of ribbon is read (i.e. finding the right event to activate the reading)

adecler:
Thanks @alelom , 

Point `(1)` is now fixed with the new commit (make sure to also update BHoM_UI)

There is nothing I can do about point `(2)`. By using an `IObject` component, you are masking the type of the actual object so all I can do is be more permissive in the list of component you can use (you should still find what you need in the search though).

kThorsager:
Hi, also using this and a small note which might be difficult to resolve, GH users are used to GHs search algorithm with it's special cases and such.
My first action was to attempt to save half a second by trying `//` which is shorthand for the panel component, but that doesn't work
image

image

Still super useful, but if this part could work it would be a super seamless transition for any user

adecler:
@kThorsager , can you add that as a new issue for feature request ? 
I'll be honest with you though, the priority for that will be veeeeery low.

kThorsager:
That's fair and sure,
Issue #497

alelom:
Another small thing that needs fixing – I realized it bugs me more than I thought:
https://github.com/BHoM/Grasshopper_Toolkit/issues/498

adecler:
@alelom , it should now be fixed. If there is nothing else, I think we should consider merging this one.

Closes #498

adecler:
Anyone wants to do a final check and approve this ?

PR #499, opened 2020-04-22
Remove the external component, but keep the reflected methods in the global menu

epignatelli:
### NOTE: Depends on
https://github.com/BHoM/BHoM_UI/pull/246

   
Closes https://github.com/BHoM/BHoM_UI/issues/223

Strategy here is to remove the External component, which is unusable in any case since the menu is too slow to display, and only allow the reflected methods in the global menu.

### Test files

This has to tested in tandem with https://github.com/BHoM/BHoM_Adapter/issues/210
You can use https://github.com/BHoM/MidasCivil_Toolkit/pull/203 as a codebase for tests.

### Additional comments

Note that the component is only in Grasshopper, and not in other UIs.
For these reflected libraries we will slowly converge to use the MachineLearning_Toolkit


adecler:
I am not sure I follow. If you remove the component
- What happens to the existing scripts that contain it ?
- What component will be created once you have selected an external method then ?
- I am not sure I like the idea of methods ONLY being available through the global menu

epignatelli:
Closing this since https://github.com/BHoM/BHoM_Adapter/issues/213#issuecomment-628574216 resorted to using another name.

PR #505, opened 2020-05-07
Take care of input type that are both list and items

adecler:

See https://github.com/BHoM/BHoM_UI/pull/253#pullrequestreview-406665005 for details


adecler:
Thanks @pawelbaran ,

Yes, this one is specifically aimed at fixing the problem @IsakNaslundBh flagged up in the UI PR referenced above. The only other case where it would help is the ModulSet from the Adapter repo but that one is never used as input. The only case where something can realistically go wrong is for an input that is an `IEnumerable/List` but I can't really see how.

IsakNaslundBh:
The issue I had is solved, just trying to think if I can break it in any way. Have not succeeded so far! 👍

PR #526, opened 2020-07-10
Adding ability to log component usage

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_UI/pull/290

   
See https://github.com/BHoM/BHoM_UI/pull/290 for details.

### DIFF
--- a/Grasshopper_UI/Templates/CallerComponent.cs
+++ b/Grasshopper_UI/Templates/CallerComponent.cs
@@ -39,6 +39,7 @@ using BH.UI.Grasshopper.Components;
 using BH.UI.Grasshopper.Others;
 using System.Collections;
 using BH.Adapter;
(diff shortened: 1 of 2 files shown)


adecler:
@FraserGreenroyd , so to be clear, the record is added when the component is ran, not when it is created. Since what we care about is the number of time a component is used (to do something) not just, for example, created on the canvas and deleted because it was incorrectly selected.

Does that changes you perception of it as an issue ? If not, I'll take a bit more time to dissect your  comment tomorrow.

FraserGreenroyd:
> Does that changes you perception of it as an issue ? If not, I'll take a bit more time to dissect your comment tomorrow.

Yes it does, I will continue running today and feedback further then - apologies as I suspect I misunderstood when the data would be captured! Will aim to feedback before your day tomorrow if I find any problems :smile:

PR #530, opened 2020-08-01
UI Refactoring

### LINKED ISSUE #466: Add ability to remove inputs with default value from any component
#### Description:
This would allow to align the current behaviour of CreateObject component from constructors for all components.

@IsakNaslundBh , @al-fisher  FYI

### LINKED ISSUE #533: Disabling and enabling a CreateCustom component causes the name of the first property to be changed to object
#### Description:
When a CreateCustom component is disabled and then re-enabled the name of the first property seems to change to object.

#### Steps to reproduce:
Step 1:
image
As expected

Step 2:
image
As expected

Step 3:
Re-enable the component
image
Warning in explode component to update outputs - not expected

Step 4:
image
After update the first property is renamed to "object"

#### Expected behaviour:
At step 3 the object names are the same as before.

#### Test file(s):
[disable issue.zip](https://github.com/BHoM/Grasshopper_Toolkit/files/5175190/disable.issue.zip)

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_UI/pull/297

   
See https://github.com/BHoM/BHoM_UI/pull/297 for details

Now leverage the changes made to the BHoM_UI, resulting is way simpler code for the components. All code is now centralised on the CallerComponent class. Only the PushCaller and the TypeCaller have a little something extra that is specific to the way Grasshopper is working.

Fixes #466
Fixes #533

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/Installers/Scripts/BuroHappold_BHoM_v3.3/0047_Grasshopper%20Toolkit/UI_Testing-Grasshopper.gh?csf=1&web=1&e=r0WguD


adecler:
All right guys, this is starting to look pretty good. 

I have spent the day testing and fixing a series of bugs created by this new version. So the major part of them should be covered. With such a large scale change, it is very likely we will have quite a few more though. 

So it would be great if you could start using Grasshopper from this branch whenever you are not on a critical project work.

IsakNaslundBh:
Some initial testing done, and some issues found on the favourite subject of versioning.

Have put a testscript [here](https://burohappold.sharepoint.com/:f:/s/BHoM/EmIILiovdtRAqKLG1c3IM3wBJCa_yOqrQy1DMCfSLYbZYA?e=TDZeKr) with some versioning of Panel creates.

Note that this requires https://github.com/BHoM/BHoM/pull/974 and https://github.com/BHoM/BHoM_Engine/pull/1942 as all those methods still exists on master. I have updated those branches to master, so should contain the required engine changes.

This script is fully functioning in terms of versioning on the master branches for the UI. Issues found here:

First three components:

image

Functioning, but missing the localX input. This can be added via the rightclick menu, so this could be acceptable, maybe even wanted, as it keeps the method functioning as before. Only worry is that the added input accepts a default value of null. If that was not the case, this could be an issue.

Several other components:

image

Has the same issue with missing added input, but above that, does not have any output. This is obviously more critical.

FraserGreenroyd:
Running this today and noticed a potential issue (in so much as, not seen this before, been running this branch, might be this branch, logging for @adecler to decide if this is an issue or not):

I have a script, using some components from Test_Toolkit, which has some explode components on the canvas which were then shortened (i.e. I clicked minus on all outputs I did not need from the explode). Opening that script, I have had to replace the explode components. All explode components are outputting `null` for each property, and right-clicking to `update outputs` jhist has `null` for all of them. If I use a new explode component it then works.

Happy to share script if needed.

IsakNaslundBh:
To follow up on @FraserGreenroyd s comment about the explode component, have also noticed an issue. Opening an "old" script (created yesterday on master branches) THe explode, without any change to it, is not working:

image

As for frasers case, replacing it with a new explode, all works fine. Also tried "UpdateOutputs" which gives the following:

image

As Fraser, can also provide script if useful.

adecler:
> Functioning, but missing the localX input. This can be added via the rightclick menu, so this could be acceptable, maybe even wanted, as it keeps the method functioning as before. Only worry is that the added input accepts a default value of null. If that was not the case, this could be an issue.

@IsakNaslundBh , yes this is intended. It basically keeps the component in the state it was when it was last used, which should be the most relevant state for the script. Not sure I understand what you mean about default value. Unselected inputs will use the default value from the method:
https://github.com/BHoM/BHoM_UI/blob/7adcdf421ec4218d10eba0926db8788f3dbf22cb/BHoM_UI/Caller/Run.cs#L131
It is worth mentioning that only optional inputs can be unselected.

adecler:
@IsakNaslundBh , the problem of missing output was coming from the fact that their name was changed in the code.  Since names are used to define what is selected, a new name that doesn't match anything was meaning it would not be selected.

Based on that, I have decided to change the selection rule for both input and output to
> Only unselect a param if it is not required and is explicitely marked as unselected in its serialised state

This applies to the inputs as well so any new input will now be selected by default. Given that you flagged the previous behaviour as a potential bug, I suppose this new one is also more intuitive anyways. 

Note that the change is made in the BhoM_UI so make sure to refresh that one before checking this bug.

IsakNaslundBh:
> It is worth mentioning that only optional inputs can be unselected.

This was basically my question. Would the localX have appeared if I had not made it optional? Think your answer is yes, and then I agree that it is fine :)

Regarding general versioning, from different versions of BHoM [this folder](https://burohappold.sharepoint.com/:f:/s/BHoM/EopxQb5d9QdDqKoD--1178kBUpVHnKPZ1WWwyI2Eg1l18Q?e=nEG9uL) could be a good resource for the testing. Tested through some other scripts in there as well with some varying issues, whilst some parts worked. Notably, the 2.3 and earlier scripts ended up doubling the inputs, something that the ones after did not.

adecler:
@IsakNaslundBh , @FraserGreenroyd , yes please. always provide a script especially when you already have one available. Makes my life so much easier to debug 😄

IsakNaslundBh:
Explode issue script in same folder: Called 'Local orientation panels to Adapters'.

Kept the script as it was before switching to this branch, so quite a few things there that are working as expected, but the issue is the explode component that should show up as red.

> @IsakNaslundBh , the problem of missing output was coming from the fact that their name was changed in the code. Since names are used to define what is selected, a new name that doesn't match anything was meaning it would not be selected.

Tested the versioning of the first script I flagged failing, and there everything seem to be working just fine now, thanks!

Would have a look at some of the older versioning scripts in the folder previously linked though. There are a few places there where some issues still occur. The 3.0 folder has an adapter component versioning script for example that is not working on my machine.

Also, in the 3.1 folder, the methods now seem to upgrade, but all wires are lost. Will double check if that is the case on master as well or not.

adecler:
I have now fixed the Explode issue that @IsakNaslundBh had flagged up. It was linked to the fact that the old version of the UI was not saving the inputs for Explode (so we ended up with an empty list used in the update). 

@FraserGreenroyd , can you check if it also fixes your problem with Explode ?

IsakNaslundBh:
After our chat arnaud, here are some more versioning issues, and the state of the scripts on the master branch as screenshots. All scripts saved [here](https://burohappold.sharepoint.com/:f:/s/BHoM/EmIILiovdtRAqKLG1c3IM3wBJCa_yOqrQy1DMCfSLYbZYA?e=tbrNej)

- 3.1 versioning script (saved in the UI testing folder). Components are upgraded correctly, but most wires are missing, script is functioning like this on master:

image


- 2.3 versioning test script (also saved in same folder), inputs are being doubled up and some wires are missing. On master one wire is missing as well, but generally works. State on master:

image

- 3.0 versioning script (adapters). Most components fail to upgrade and many wires are missing. On master functioning like this (had to re-plug the toggle to the pull for the nodes to be output in the pull component to the bottom right): 

image

Guess this last case can be an interesting one as it is pointing to methods in the UI, wonder if any of those have changed?

adecler:
@IsakNaslundBh , I have done my best to fix the issue with components from 2.3 but there are cases where I just don't have the information needed to fix it. Basically, it will work if the inputs haven't changed. If they have, the trick of toggling off the new input with the selection menu solves it but I'd rather advise the user to create a new component at this point. 

So I have updated the error message to tell the user what is going on and which component to create. Let me know if that is sufficient for you.

IsakNaslundBh:
> @IsakNaslundBh , I have done my best to fix the issue with components from 2.3 but there are cases where I just don't have the information needed to fix it. Basically, it will work if the inputs haven't changed. If they have, the trick of toggling off the new input with the selection menu solves it but I'd rather advise the user to create a new component at this point.
>
>So I have updated the error message to tell the user what is going on and which component to create. Let me know if that is sufficient for you.

Thanks, will have a look.


Can also mention, I have added a smaller test script, exploding a coordinatesystem that could be used to test the explode component. Significantly smaller than the larger script in the folder and showcases the same behaviour.

IsakNaslundBh:
> > @IsakNaslundBh , I have done my best to fix the issue with components from 2.3 but there are cases where I just don't have the information needed to fix it. Basically, it will work if the inputs haven't changed. If they have, the trick of toggling off the new input with the selection menu solves it but I'd rather advise the user to create a new component at this point.
> > So I have updated the error message to tell the user what is going on and which component to create. Let me know if that is sufficient for you.
> 
> Thanks, will have a look.

Refactoring branches, with no change to the panel create (master on BHoM and Engine). Works better than before! :)

image

Master UI _with_ the changes to the panel create method:

image

Refactoring branches on UI _with_ the changes to the panel create method:

image

I do not get the error message you are referring to, but rather the "This is a magic component" message. If it would be possible to make sure that the error message is coming through, and that the component can be fixed I think that would be awsome, but at least, it does not look like a step backwards from the tests. Rather a step forward as it works better for the not changed component, and is at least going orange when not functioning (compared to the same on master where the component is grey but malfunctioning)

FraserGreenroyd:
Explode still has an issue for me.

I had to remake my script cause I'd saved it broken yesterday by mistake when I was using it. Huzzah for muscle memory to save scripts... To do this I installed today's (27th) daily alpha, remade my script, then recompiled BHoM_UI and GH_TK over the top of that alpha. This also includes the engine changes from yesterday which were merged.

On reopening my script, my explode components only giving me `null` out. This explode was deliberately cut down previously to only explode the name (as that's the only property I care about at this stage of the script), so not sure if that affects it (due to selected/unselected outputs). See image below for the output.

I've uploaded the script in question. Please note, it is using components from Test_Toolkit, as it is the compliance checking script, so you may need to compile the latest version of that to get any DLLs from there as necessary. Test_Toolkit has postbuilds set up so that should work with our ProgramData stuff - it does for me at least to compile it on the top of the alphas :smile:

[Script link](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_UI/BHoM_UI-FrameworkRefactoring/FG%20-%20Explode%20Issue%20from%20Compliance%20Script.gh?csf=1&web=1&e=3fcA6e)

image

adecler:
@IsakNaslundBh , from opening the file you gave me (2.3 versioning issues) and using the PR branches for oM and Engine 👍 

image

I am really confused as why we would have different results.

adecler:
Thanks @FraserGreenroyd , I'll have a look. 
Note that the Test_Toolkit from master was not compiling for me, even after running the latest installer. 
I have raised an issue [here](https://github.com/BHoM/Test_Toolkit/issues/271) with the changes required to fix it.

adecler:
Ok, guys, old Explode components should now deserialise correctly. 
I really need to have a chat with the guy that thought that using a different serialisation system for Explode was a good idea :wink:

adecler:
Regarding the wires disappearing, it was actually a bug that was already there in the master version. Just hidden better because the adapter components were not even upgraded at all. 

The existing bug is that we are relying on Param.Recipients to reconnect the wires BUT that list is always empty when the file is loaded. So updates on inputs would reconnect correctly but not update on outputs.  I'll see what I can do to find that information in another way or just do the rewiring as a second phase.

IsakNaslundBh:
> @IsakNaslundBh , from opening the file you gave me (2.3 versioning issues) and using the PR branches for oM and Engine
> 
> 
> I am really confused as why we would have different results.

Getting the same as you do now. Must have missed compiling the versioning toolkit or something. Now works/looks good, Better error-message than before, and like the possibility to fix the component by changing the inputs!

adecler:
Alright, missing link should be fixed. 

Side note, you gotta love those little Easter eggs:

image

FraserGreenroyd:
So explode is now working 🥳 however I have noticed that when I right click and wish to add/remove outputs, it says add/remove inputs instead. See image. Just a minor wording change needed probably!

image

adecler:
Wow, that previous serialisation of the Explode output parameters was really sloppy. It was actually saving the parameters as inputs.

adecler:
I think I have now fixed all the bug that were reported here (most of them fixed in the BHoM_UI). 
Let me know if you disagree.

IsakNaslundBh:
Found some further issues with the explode, especially the custom data. Have saved a script named "Explode issues custom data" in the folder.

image

Component asks you to update the outputs straight away after plugging it in. Doing so gives no result.

adecler:
@IsakNaslundBh , that bug is an old one that is on the master too if I believe the comment in the code:

image

testing in GH, that seems to confirm it:

image

I was a bit scared of doing a fix that might break a lot of components as most of them are mono-output. So I fixed the Explode.Run() method instead by making it return an object instead of a list in case of a single output.

IsakNaslundBh:
> I was a bit scared of doing a fix that might break a lot of components as most of them are mono-output. So I fixed the Explode.Run() method instead by making it return an object instead of a list in case of a single output.

Makes sense, and seem to be working now!

al-fisher:
@adecler - have discovered issues with selecting and then deselecting inputs to components. 

This can create both duplicate inputs - as well as inputs that are either not actually "active", or mislabelled. i.e assigning `10` to `Y` as an input is not giving the correct output.

AddAndRemoveInputs

image
image

adecler:
Nice catch @al-fisher ! Now fixed in the BHoM_UI PR.
It's pretty crazy that I lasted that long without seeing the bug a single time 😕

al-fisher (approved):
Thanks @adecler - adding and removing outputs bug-fix in BHoM_UI is now working as expected for me.

IsakNaslundBh (approved):
Putting as a co-approval of https://github.com/BHoM/BHoM_UI/pull/297#pullrequestreview-485693306

Niggly bits fixed and seem to be working as expected!

PR #535, opened 2020-09-11
Usage log now saving caller type name instead of caller name

### LINKED ISSUE #534: Analytics should log caller type name instead of caller name
#### Description:
`Caller.Name` gives us the name of the component as it is seen in the UI while `Caller.GetType().Name` gives us the type of component, which is more relevant in this case.

adecler:

Closes #534

Same as for other UIs: Caller Type name should be saved instead of caller name in usage log.


adecler:
Quick reminder that this is awaiting review. Let me know if you don't have the time.

al-fisher (approved):
LGTM.
Can confirm now CallerName is storing Caller Type name now.

For this example:
image



On master:
`"CallerName" : "CreateObject"`

This PR
`"CallerName" : "CreateObjectCaller"`

PR #546, opened 2020-09-22
Fix error messages spreading to BHoMParam components

### LINKED ISSUE #545: Grasshopper_Toolkit: Errors from components show up on Parameters
#### Description:

Error coming from unrelated components can sometimes show up on grasshopper parameters.

The case found so far is when the IObejct blocks the preview due to to many objects on the param:

image

#### Steps to reproduce:

Make IObejct param throw a message, while at the same time you have an error elsewhere in the script

#### Expected behaviour:

For error from the param itself to be the only ones seen on the parameter

#### Test file(s):

https://burohappold.sharepoint.com/:f:/s/BHoM/EoBU_y7aC8BAomZexcM9h6QBJZnSAqO0zzboqljNf9GuFQ?e=ay8uLg

adecler:

Closes #545 

The BHoMParam components don't have a clear entry point to clear pre-existing event messages. This leads to situations like shown in issue #545. 

The simplest solution seems to clear the events once they have been pushed to Grasshopper. I have left that as an optional argument to the `ShowEvents` method just in case there would be exceptional cases. 

### Test files
Test file provided in the issue.

### DIFF
--- a/Grasshopper_UI/Helpers/Logging.cs
+++ b/Grasshopper_UI/Helpers/Logging.cs
@@ -37,7 +37,7 @@ namespace BH.UI.Grasshopper
         /**** Public Methods              ****/
         /*************************************/
 
-        public static void ShowEvents(GH_ActiveObject component, List<Event> events)
+        public static void ShowEvents(GH_ActiveObject component, List<Event> events, bool clearShownEvents = true)
         {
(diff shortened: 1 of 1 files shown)


IsakNaslundBh (approved):
LGTM, works!

PR #547, opened 2020-09-22
Fix added param input/output when already exists

### LINKED ISSUE #537: Grasshopper should just switch to replace when asked to add an input that already exists.
#### Description:
See https://github.com/BHoM/BHoM_UI/pull/303 and https://github.com/BHoM/BHoM_UI/pull/304 for reference

adecler:

Closes #537 

In the rare instance when a component was saved incorrectly (from previous version of the code), we might run the risk of having duplicated inputs/outputs. The situation occurs when a param is on the Grasshopper component but not saved by BHoM. So when we reopen the file, the BHoM_UI will ask for that input/output to be added again.

This PR fixes that issues by making sure we delete the old GH param after having moved the wires across to the new one. So it's very similar to an update with the difference that the position of the param is dictated by the the info in `ParamAdded` instead of the location of the old GH param. 

@LMarkowski, added you as review since you had raised the original issue in https://github.com/BHoM/BHoM_UI/pull/304#pullrequestreview-489394179

@BingWangUS , added you since you raised similar concerns in Teams

@IsakNaslundBh , added you just so you can review the code itself. It's a fairly straightforward change but better safe than sorry 😄

### DIFF
--- a/Grasshopper_UI/CallerComponent/OncallerModified.cs
+++ b/Grasshopper_UI/CallerComponent/OncallerModified.cs
@@ -88,7 +88,18 @@ namespace BH.UI.Grasshopper.Templates
 
         protected virtual void UpdateInput(ParamAdded update)
         {
-            Params.RegisterInputParam(update.Param.ToGH_Param(), update.Index);
+            IGH_Param newParam = update.Param.ToGH_Param();
+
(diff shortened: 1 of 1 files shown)


BingWangUS (approved):
I can confirm it resolve the issue I experienced.

PR #550, opened 2020-09-29
Add ability to bake BHoMObjects and BHoMGeometry

### LINKED ISSUE #549: Add ability to Bake BHoMGeometry as well as BHoMObejcts
#### Description:

Adding Feature to be able to Bake BHoMGeometry.

Also, adding feature to Bake any BHoMObject that has a `Geometry()` method implemented (any Previewable object). Can then assign the name of the object (if any) to the baked obejct

IsakNaslundBh:

Closes #549

Did some quick digging after the workshop, and realised this was relatively simple, and actually already half-solved for geometry, so went ahead and just fixed it. Hope this does not derail anything @adecler , if so, please feel free to close/park.

- Making it possible to bake out BHoMGeometry
   - One half of this was actually already working. You actually were able to bake BHoMGeometry from a Data parameter. For this, Just needed to make the BHoMGeometry parameter bake aware as well.
- Adding Baking capability to the BHoMObject Goo and paramter
- Adding Baking capability to the IObjectGoo and paramter

### Test files

Try baking some random BHoMObejcts and geometries. Should bake any BHoMObejct that is previewable, no others.

### DIFF
--- a/Grasshopper_UI/Components/Parameters/Param_BHoMGeometry.cs
+++ b/Grasshopper_UI/Components/Parameters/Param_BHoMGeometry.cs
@@ -22,11 +22,15 @@
 
 using BH.UI.Grasshopper.Properties;
 using BH.UI.Grasshopper.Templates;
+using Grasshopper.Kernel;
 using System;
+using Rhino;
+using Rhino.DocObjects;
+using System.Collections.Generic;
 
 namespace BH.UI.Grasshopper.Parameters
 {
-    public class Param_BHoMGeometry : BHoMParam<Engine.Grasshopper.Objects.GH_IBHoMGeometry>
+    public class Param_BHoMGeometry : BHoMParam<Engine.Grasshopper.Objects.GH_IBHoMGeometry>, IGH_BakeAwareObject
     {
         /*******************************************/
         /**** Properties                        ****/
@@ -38,15 +42,42 @@ namespace BH.UI.Grasshopper.Parameters
 
         public override string TypeName { get; } = "BHoM Geometry";
 
+        public bool IsBakeCapable { get; } = true;
+
+        /***************************************************/
+        /**** IGH_BakeAwareObject methods               ****/
+        /***************************************************/
+
+        public void BakeGeometry(RhinoDoc doc, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_IBHoMGeometry item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, null, out guid))
+                    obj_ids.Add(guid);
+            }
+        }
+
+        /*******************************************/
+
+        public void BakeGeometry(RhinoDoc doc, ObjectAttributes att, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_IBHoMGeometry item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, att, out guid))
+                    obj_ids.Add(guid);
+            }
+        }
 
         /*******************************************/
         /**** Constructors                      ****/
         /*******************************************/
 
-        public Param_BHoMGeometry(): base("BHoM geometry", "BHoMGeo", "Represents a collection of generic BHoM geometries", "Params", "Geometry")
+        public Param_BHoMGeometry() : base("BHoM geometry", "BHoMGeo", "Represents a collection of generic BHoM geometries", "Params", "Geometry")
         {
         }
-        
+
         /*******************************************/
     }
 }
--- a/Grasshopper_UI/Components/Parameters/Param_BHoMObject.cs
+++ b/Grasshopper_UI/Components/Parameters/Param_BHoMObject.cs
@@ -22,11 +22,15 @@
 
 using BH.UI.Grasshopper.Properties;
 using BH.UI.Grasshopper.Templates;
+using Grasshopper.Kernel;
 using System;
+using Rhino;
+using Rhino.DocObjects;
+using System.Collections.Generic;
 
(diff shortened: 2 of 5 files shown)


Grasshopper_UI/Components/Parameters/Param_BHoMGeometry.cs
```diff
@@ -38,15 +42,42 @@ public class Param_BHoMGeometry : BHoMParam<Engine.Grasshopper.Objects.GH_IBHoMG
...
+        }
+
+        /*******************************************/
+
+        public void BakeGeometry(RhinoDoc doc, ObjectAttributes att, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_IBHoMGeometry item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, att, out guid))
```
adecler, line 68: Probably a good idea to check that `item` is not null

Grasshopper_UI/Components/Parameters/Param_BHoMGeometry.cs
```diff
@@ -38,15 +42,42 @@ public class Param_BHoMGeometry : BHoMParam<Engine.Grasshopper.Objects.GH_IBHoMG
...
+        /***************************************************/
+        /**** IGH_BakeAwareObject methods               ****/
+        /***************************************************/
+
+        public void BakeGeometry(RhinoDoc doc, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_IBHoMGeometry item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, null, out guid))
```
adecler, line 56: Probably a good idea to check that `item` is not null

Grasshopper_UI/Components/Parameters/Param_BHoMObject.cs
```diff
@@ -38,6 +42,33 @@ public class Param_BHoMObject : BHoMParam<Engine.Grasshopper.Objects.GH_BHoMObje
...
+        /***************************************************/
+        /**** IGH_BakeAwareObject methods               ****/
+        /***************************************************/
+
+        public void BakeGeometry(RhinoDoc doc, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_BHoMObject item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, null, out guid))
```
adecler, line 56: Probably a good idea to check that `item` is not null

Grasshopper_UI/Components/Parameters/Param_BHoMObject.cs
```diff
@@ -38,6 +42,33 @@ public class Param_BHoMObject : BHoMParam<Engine.Grasshopper.Objects.GH_BHoMObje
...
+        }
+
+        /*******************************************/
+
+        public void BakeGeometry(RhinoDoc doc, ObjectAttributes att, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_BHoMObject item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, att, out guid))
```
adecler, line 68: Probably a good idea to check that `item` is not null

Grasshopper_UI/Components/Parameters/Param_IObject.cs
```diff
@@ -58,6 +62,33 @@ public override bool IsPreviewCapable
...
+        /***************************************************/
+        /**** IGH_BakeAwareObject methods               ****/
+        /***************************************************/
+
+        public void BakeGeometry(RhinoDoc doc, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_IObject item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, null, out guid))
```
adecler, line 76: Probably a good idea to check that `item` is not null

Grasshopper_UI/Components/Parameters/Param_IObject.cs
```diff
@@ -58,6 +62,33 @@ public override bool IsPreviewCapable
...
+        }
+
+        /*******************************************/
+
+        public void BakeGeometry(RhinoDoc doc, ObjectAttributes att, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_IObject item in this.VolatileData.AllData(true))
+            {
+                Guid guid;
+                if (item.BakeGeometry(doc, att, out guid))
```
adecler, line 88: Probably a good idea to check that `item` is not null

adecler (changes requested):
Brilliant! Thanks @IsakNaslundBh 😄 

Tested both on output param and param components and it works great for the 3 categories (Geometry, IObject & BHoM Object). 

I have never really used bake myself so probably a good idea for someone else to review as well. Otherwise, looks good to me both on code an functionality side with one minor caveat.

Grasshopper_UI/Components/Parameters/Param_BHoMGeometry.cs
IsakNaslundBh: the 'true' in `this.VolatileData.AllData(true)` is supposed to take care of this

image

but can add an extra null check anyway, just to be extra safe if you think that is worth it? :)

adecler: Ha! Fair enough 😋 

Happy to approve then.

adecler (approved):

pawelbaran (approved):
LGTM, works as expected, the only issue I see is the fact that pulled objects need to be cast to `BHoMObject` before they can be baked - this, however, seems to be out of the scope of this PR therefore happy to approve.

Issue #540, opened 2020-09-17
Grasshopper_Toolkit: GetProperty returns unconverted BHoM geometries


pawelbaran:
#### Description:

The output of `GetProperty` comes out as BHoM geometry, which means that it needs to be explicitly converted before it starts to behave like a native GH/Rhino geometry:

![image](https://user-images.githubusercontent.com/26874773/93459415-f6ca3b00-f8e1-11ea-9e0f-53e2c033e7f4.png)

It is a quite annoying issue, particularly to the non-expert users.

#### Test file(s):

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FGrasshopper%5FToolkit%2FGrasshopper%5FToolkit%2DIssue540%2DGetPropertyCast&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

adecler:
Ok, I haven't touched that part of the code in a very long time so I am a bit rusty in the specifics of GH. @epignatelli , feel free to jump in if I am missing the obvious. In the meantime, I will document what  I find here.

Any Grasshopper Param (including the BHoM parameters we created) has two casting methods:
- `CastFrom(object source)` :  this is called when the parameter is reading its input value (i.e. when the component that own the parameter is being run).
- `CastTo<Q>(ref Q target)` this is called when the parameter is being asked to provide it value to another parameter (i.e. when the param is plugged into the input of another component and that component is executed).

This is a great feature from GH because it means we don't have to constantly convert back and forth between BHoM and Grasshopper when we are chaining BHoM components (we can just stay in the BHoM realm). So big improvement in efficiency an reduction of risk of errors due to casting (e.g. nurbs curves)

Here's  where the problem lies in this case: for some reason the GH Area component is not asking the PlanarSurface output param to cast its content. As we can see, it works fine for things like the GH component that calculate the distance between points.

![image](https://user-images.githubusercontent.com/16853390/93553300-3c037100-f9a5-11ea-97a3-3e0e8a80a61f.png)

Since the input of the `Area` component is a `Geometry` param, I have tried this:

![image](https://user-images.githubusercontent.com/16853390/93553470-a3b9bc00-f9a5-11ea-8ed6-72a1ae04a54e.png)

[...]

Issue #529, opened 2020-07-28
CreateCustom does not convert breps


LMarkowski:
#### Description:

While creating `CustomObject` with `CreateCustom` rhino geometry seems not to be converted to any `BHoMObject` but stored in the object properties as rhino object instead.

#### Steps to reproduce:

Create `CustomObject` from a brep.

#### Expected behaviour:

Should convert brep to BHoM Geometry.

#### Test file(s):

[Test file](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FGrasshopper%5FToolkit%2FGrasshopper%5FToolkit%2DIssue529%2DCreateCustomDoesntConvert)

LMarkowski:
@al-fisher fyi

adecler:
Closing as replaced by #543

PR #557, opened 2020-10-22
Replace all Param_ScriptVariable with our own solution

### LINKED ISSUE #543: Replace all Param_ScriptVariable with our own solution
#### Description:
This is required to get access to `CastFrom(object source)` and `CastTo<Q>(ref Q target)` methods.
This will solve the problem of Rhino geometry not being casted back and forth (e.g. `CreateCustom` & `GetProperty` components).

Related issues:
- #540: GetProperty returns unconverted BHoM geometries
- #529: CreateCustom does not convert breps

This will have to be done for 4.0 instead of 3.3 since this represent a change that can potentially break the components using `Param_ScriptVariable`.

adecler:

Closes #543

This is required to get access to `CastFrom(object source)` and `CastTo<Q>(ref Q target)` methods.
This will solve the problem of Rhino geometry not being casted back and forth (e.g. `CreateCustom` & `GetProperty` components).

I also had to do a bit of refactoring of the BH parameters as that addition of the variable parameter would have added to already existing code duplication. So now, the `IObject`, `BHoMObject`, and `Variable` parameters are all inheriting from the `Bakeable` parameter. I would have liked to also add the `Geometry` parameter to that list but it was already inheriting from a different class than the others and it was out of the scope of this PR to change that. 

### Test files
- Make sure that those two issues are resolved:
   - #540: `GetProperty` returns unconverted BHoM geometries
   - #529: `CreateCustom` does not convert breps

- In general, test that were relying on variable parameters such as `CreateCustom`, `SetProperty` and `Explode`. 
- Also make sure that old scripts containing those are still working


IsakNaslundBh:
As discussed with @adecler think it is sensible to keep the goos to object for this PR and resolve it in a separate issue. Issue raised here:

https://github.com/BHoM/Grasshopper_Toolkit/pull/557

Will test this with the expectation that the previous params all work as before, and any param that previously returned script variable not should have the added benefit this PR actually set out to resolve!

Grasshopper_UI/Templates/BakeableParam.cs
```diff
@@ -0,0 +1,101 @@
...
+using Grasshopper.Kernel;
+using System;
+using Rhino;
+using Rhino.DocObjects;
+using System.Collections.Generic;
+using Grasshopper.Kernel.Types;
+
+namespace BH.UI.Grasshopper.Templates
+{
+    public abstract class BakeableParam<T> : BHoMParam<T>, IGH_BakeAwareObject where T : class, IGH_Goo
```
IsakNaslundBh, line 35:
```suggestion
    public abstract class BakeableParam<T> : BHoMParam<T>, IGH_BakeAwareObject where T : class, IGH_Goo, IGH_BakeAwareData
```

Grasshopper_UI/Templates/BakeableParam.cs
```diff
@@ -0,0 +1,101 @@
...
+        public bool IsBakeCapable { get; } = true;
+
+
+        /***************************************************/
+        /**** IGH_BakeAwareObject methods               ****/
+        /***************************************************/
+
+        public void BakeGeometry(RhinoDoc doc, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_BakeableObject<T> item in this.VolatileData.AllData(true))
```
IsakNaslundBh, line 69:
```suggestion
            foreach (T item in this.VolatileData.AllData(true))
```

The T on the param level here is already a Goo, so should be just T, not `....Goo<T>`

It required the additional constraint above as well to work.

Layercake GH.....

Grasshopper_UI/Templates/BakeableParam.cs
```diff
@@ -0,0 +1,101 @@
...
+                if (item.BakeGeometry(doc, null, out guid))
+                    obj_ids.Add(guid);
+            }
+        }
+
+        /*******************************************/
+
+        public void BakeGeometry(RhinoDoc doc, ObjectAttributes att, List<Guid> obj_ids)
+        {
+            foreach (Engine.Grasshopper.Objects.GH_BakeableObject<T> item in this.VolatileData.AllData(true))
```
IsakNaslundBh, line 81:
```suggestion
            foreach (T item in this.VolatileData.AllData(true))
```

IsakNaslundBh (changes requested):
Some tweaks needed to get baking working. Suggested changes below seem to do it.

Grasshopper_UI/Templates/BakeableParam.cs
IsakNaslundBh: Instead of changing to T could instead had changed to `IGH_BakeAwareData` I think

IsakNaslundBh:
Another thing found that I am not sure if it is out of scope of the PR or not is that any component placed before this change does not seem to reap the quite massive benefit this PR brings. Not sure if that is out of scope for this PR or not!

Can also add, to not sound all negative, I _LOVE_ this update. Think it will reduce so much of annoying friction we have been experiencing for various things!

IsakNaslundBh:
Some issues exploding/get proeprty from obejcts with list of list proeprty. Not sure if it is related to this PR (do not think it is) in any shape or form, but:

image

Have raised https://github.com/BHoM/Grasshopper_Toolkit/issues/559 for it if better of tackle it separately. Also have a testfile in there.

Grasshopper_UI/Templates/BakeableParam.cs
adecler: That interface is for the level of the Goo, not the params

Grasshopper_UI/Templates/BakeableParam.cs
adecler: > Instead of changing to T could instead had changed to `IGH_BakeAwareData` I think, but can leave it up to you @adecler

Yes, this is more correct as the interface `IGH_Goo` (which is the constraint on `T`) doesn't provide a `BakeGeometry` method

Grasshopper_UI/Templates/BakeableParam.cs
adecler: I don't believe this is correct. See comment above

adecler:
Thanks @IsakNaslundBh ,

I have fixed the baking issue although made a slightly different correction than yours.

As for your other two comments:

>Another thing found that I am not sure if it is out of scope of the PR or not is that any component placed before this change does not seem to reap the quite massive benefit this PR brings. Not sure if that is out of scope for this PR or not!

Yes, I am aware of that but decided it was not worth the risk and the decent amount of work required to make that happen (you basically have to individually replace every script parameter and recreate the connections). The old scripts were working fine without that feature and will still work fine despite the change. That's plenty enough in my book 😄 

>Some issues exploding/get property from objects with list of list property. Not sure if it is related to this PR (do not think it is) in any shape or form

Yes, this is a separate issue that already happens on the master.  Note that getting a lists of lists property works fine if you extract that property from a single object. The error occurs when you have multiple objects in the first input. Will need to investigate further.

image

Grasshopper_UI/Templates/BakeableParam.cs
IsakNaslundBh: yes, hence putting it on the where part, as in `T: ... , ... ,   IGH_BakeAwareData` :)

Grasshopper_UI/Templates/BakeableParam.cs
IsakNaslundBh: Yeah, the the `T` solution would require the additional constraint on the `T` as shown above :)

Think we probably should put that in anyway, as if we want the Param to be bakeable, then the Goo should be as well (whether we use the `T` or the explicit interface solution matters less I think)

IsakNaslundBh (changes requested):
Found an issue with the casting to valuetypes, in particular, to numbers:

image


Will need to ensure that the CastTo for the new generic variable is able to be casted to any of the base level types such as integers and doubles when appropriate.

adecler:
Thanks Isak,

This should now be fixed for all primitives. I have also added support for Guids:

image

IsakNaslundBh (approved):
This seem to be working well for me now. Would not mind a second review as well @al-fisher @pawelbaran @alelom as this is touching a lot of stuff, but approving to unlock @adecler

alelom (dismissed):
Working well for me too.

IsakNaslundBh (changes requested):
Testing this a bit more, seems like the issue with primitives presists when used as inputs, even though the parameters are working fine:

image

IsakNaslundBh (changes requested):
Found another issue now.


Can not covert from the new generic goo to our type goo.

image


Think the issue for this is that the casting in the Type goo needs to be expanded:

image

adecler:
QuickCast ? What now ? 😕 

I was getting a bit tired of this casting system from DR so I creted a separate `Compute` file with castings for all the existing GH_Goo (that had a corresponding `ToGHX` method). That should take care of it. The casting method taking care of the `IGH_QuickCast` ones covers all the possible types I found through reflection so that should now be good too.

image

I have also taken care of the error we had on the casting in the `Type` Goo. 

Hopefully that should now be enough to settle that issue for good 🤞

FraserGreenroyd:
I've removed my request for review because I am heavily out of the loop on this one (apologies) and it appears that @IsakNaslundBh and @alelom are on track with this. Please feel free to re-request if my review is needed, but I feel the others have it in hand :smile:

adecler:
@IsakNaslundBh , would be able to do another review (hopefully the final one) ? 
This repo is getting pretty crowded with another PR still waiting to be pushed and more issue to fix this sprint.

IsakNaslundBh (approved):
Found another case not working, but was not working on master either (using addition with vectors):
image

On master it is behaving as the geometry param is (in the explode) which is arguably worse as it does not give any errors, while the new one does, so not enough to hold of this PR.

Given the overly complex casting system of GH, I am not sure if I have missed to check something, but look to me like the core principles are working as expected now at least! Could preform the basic operations with numbers, see geometry and bake straight from the pull etc!

Ready to approve now for further testing by more users in the alphas!

adecler:
Thanks @IsakNaslundBh ,  I know where this one is coming from: 

In `PreferredCast`, I treat any `IObject` differently from primitive types since the later don't need baking and other fancy Casting based on what they are plugged into. Sadly Rhino Vectors and Points are structures not classes. So they are treated by GH the same way as primitives (with that parallel messy casting system). There might be a way to keep the best of both world but I am not sure how. So for now, I decided it was a small price to pay. Can you raise that as a separate issue though ? When I can find the courage, I can have another look into it.

IsakNaslundBh:
Issue already raised here: https://github.com/BHoM/Grasshopper_Toolkit/issues/565 😄

PR #564, opened 2020-11-11
Grasshopper toolkit: Keeping flattening settings when replacing an input/output

### LINKED ISSUE #562: Grasshopper_Toolkit: Flatten disappears in copy pasting components
#### Description:

Found a bug where the flatten disappears from the push component after copy paste. Marking as critical as this can impact save and open as well, as they are using the same systems:

Flatten gone

#### Steps to reproduce:

Copy paste a push component. Note that this only happens if you have something plugged in to the push. If left empty, the flatten stays.

#### Expected behaviour:

For any flatten/graft/similar to stay on the component after copy paste and save and re-open.

#### Test file(s):

Can not provide one as the problem affects the save and open system

adecler:

Closes #562

See the issue for more details.

This was actually only happening when the input/output needed an upgrade. When replacing it, the links were copied over but not the flattening settings. 

### Test files
Reproduce the example provided in the issue and make sure it now works as expected.

### DIFF
--- a/Grasshopper_UI/CallerComponent/OncallerModified.cs
+++ b/Grasshopper_UI/CallerComponent/OncallerModified.cs
@@ -122,6 +122,8 @@ namespace BH.UI.Grasshopper.Templates
                 IGH_Param newParam = update.Param.ToGH_Param();
 
                 MoveLinks(oldParam, newParam);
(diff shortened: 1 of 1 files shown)


IsakNaslundBh (approved):
LGTM

PR #567, opened 2020-11-13
Grasshopper_Toolkit: Refactor the code around BHoM Goos

### LINKED ISSUE #364: Grasshopper_Toolkit: GH_IBHoMGeometry should specify  GH_GeometricGoo<IGeometry>
Instead of  `GH_GeometricGoo<object>`.

We have now separated the rhino entity and the bhom entity in the Geometry.

`Value` and `m_Value` always refer to a `BH.oM.Geometry.IGeometry` object
`m_RhinoGeometry` works in the background and always refers to a `Rhino.Geometry` object

### LINKED ISSUE #365: Grasshopper_Toolkit: GH_BHoMObject should specify GH_BHoMGoo<BHoMObject>
instead of GH_BHoMGoo<object>.

Following the amendement that brought to #364.
At the moment you can pass anything in `GH_BHoMObject` which is conflicting with its base definition. The reflection here is not symmetric, and not that transparent.
e.g. `IGeometry` is not a `BHoMObject` and should fail, whereas it currently succeeds.

Passing a `IGeometry` would work in `GH_IObject` though.

### LINKED ISSUE #406: ToString() method should use the Engine.Reflection.ToText method
#### Description:

For each `Goo`, the `ToString()` method should use the `Engine.Reflection.ToText` method to enhance the way we display objects in panels.

For example, a Rhino point is displayed in a panel as `{x, y, z}`. We should do the same for the BHoM.

### LINKED ISSUE #415: Move Compute.Render methods from Engine to UI
#### Broken rules:
Render methods are interacting with the UI and therefore should not be in the Engine

### LINKED ISSUE #544: Move the logic inside `GH_IBHoMGeometry.CastFrom()` to a new IFromRhino(object obj) method in Rhinoceros_Toolkit
#### Broken rules:

While testing the Rhinoceros_Toolkit, I realised we cannot do this currently:
image

This is because Rhino does not has a parent class for all of its geometries (the highest abstraction it has is `GeometryBase` e.g. a `Point` is not a `GeometryBase`, while all solids are). Effectively, the highest parent class to Rhino's geometries is `System.Object`.

However, we can do this:
image

This is because we have this logic: https://github.com/BHoM/Grasshopper_Toolkit/blob/97068c716ce0690fd0f1057e1ac6898a47039fc5/Grasshopper_Engine/Objects/Types/GH_IBHoMGeometry.cs#L200-L231

which needed in order to make the Grasshopper Goos work.

However (after a discussion with @al-fisher and @IsakNaslundBh) we realized that the logic should belong to a `IFromRhino(object obj)` method in Rhinoceros_Toolkit, because this is really a "Convert" logic belonging to Rhino, not Grasshopper. The Grasshopper `GH_IBHoMGeometry.CastFrom()` method should call Rhino's `IFromRhino(object obj)`.

#### Suggestions to restore compliance:

Move the logic inside `GH_IBHoMGeometry.CastFrom()` to a new IFromRhino(object obj) method in Rhinoceros_Toolkit.

### LINKED ISSUE #558: Grasshopper_Toolkit: BHoM Goos and Params should be more restrictive
#### Description:

Slightly related to refactoring done here: https://github.com/BHoM/Grasshopper_Toolkit/pull/557 but not fully.

We should update the requirements on the 

- `GH_BHoMObject`
- `GH_BHoMAdapter`
- `GH_IBHoMGeometry`
- `GH_IObject`

To be restricted to only handle the types they are actually supposed to be handling.

Right now all of the above are set to accept `System.Object` which means they are basically set up to handle anything. We should make better use of Grasshopper built in system for protecting against wrong inputs and make sure you actually use the correct params when needed, and make sure any potential errors are correctly flagged by the UI.

To be able to do this we should change:

- `GH_BHoMAdapter : GH_BHoMGoo<IBHoMAdapter>`
- `GH_BHoMObject : GH_BakeableObject<BHoMObject>`
- `GH_IBHoMGeometry : GH_GeometricGoo<IGeometry>`
- `GH_IObject : GH_BakeableObject<IObject>`

More than this, we will also need to update any place in the code of any of the base classes that are referring to `object` on the basis that that is the expected type to `T`.

It will also be important that any `CastTo` method that fails to convert the object to the expected type returns `false` which will lead to the GH UI to prompt its standard message, such as:

image

Whilst doing this, a general overhaul of the whole casting mechanism might be required, as it is currently done diffrently for the geometry param to the IObejct param, even if both can, and should, be able to handle IGeometry.

Might also require an update to the `IFromGoo<T>` methods to return a `bool` for success and make use of an `out` parameter for the converted type instead, to avoid having to rely on letting the method crash and catching the exception.

Another part that will need to be looked at is this part in the BakeableParam:

```
        public override bool IsPreviewCapable
        {
            get
            {
                if (VolatileDataCount < m_MaxItemsPreview || m_ForcePreview && !Hidden)
                {
                    // Setting IsPreviewCapable from true to false clears the grasshopper geometry cache
                    // be mindful to use m_ForcePreview and Hidden variables only in case the number of objects
                    // to display is actually high
                    this.ClearRuntimeMessages();
                    return base.IsPreviewCapable;
                }
                Engine.Reflection.Compute.RecordNote("Preview has been disabled to prevent a slowdown due to the high number of objects." +
                    "Right click and set the a new items limit to force the preview at your own risk.");
                Helpers.ShowEvents(this, Engine.Reflection.Query.CurrentEvents());
                return false;
            }
        }
```

as some initial tests I made seem to indicate that this hides any error messages produced by GH. Would investigate if the `                    this.ClearRuntimeMessages()` is actually needed and would change the record note part making use of our system into straight calls to the `AddRuntimeMessage` as this is extremely tied to GH and does not need to pass through our logging system.

Bit of wall of text here, but as discussed with @adecler putting everything found so far in this issue for further reference!

adecler:
### Depends on:

https://github.com/BHoM/BHoM_Engine/pull/2163
https://github.com/BHoM/Rhinoceros_Toolkit/pull/178

First 3 commits:
Closes #415

Following commits:
Closes #558
Closes #364 
Closes #365 
Closes #406
Closes #544 

- First 3 commits prepare the repo for the changes required by #558. More specifically, BHoM_Adapter cannot be referenced in the Engine. Since the Goos are specific to the GH UI and have nothing to do with an Engine, that was the final reason needed to do the migration. 

- Following commits are solving #558 and all the issues that gravitate around it.

### Test files
Make sure that input/output params are still working correctly on existing scripts and new one. The main focus is to test things like conversions, previews, and baking

### Additional comments
Functionality in the first 3 commits hasn't been touched in any way. This was purely limited to the minimum amount of changes required to migrate the files from Engine to UI.


adecler:
I have now pushed the changes required to fix #558 to unblock things on my side. This makes things a bit more scary to review given that large amount of changes created in the first 3 commits. Those 3 are purely moving things around though without any functionality changes so I suggest you have a look at each commit separately. 

I have updated the title and first comment to reflect this.

IsakNaslundBh (changes requested):
Have not had time to dig in deeper as to why this is happening, but trying to open the script in this folder:

https://burohappold.sharepoint.com/:f:/s/BHoM/Eha-mq-0l55KryMeBe543YkBf2M6CQUaOaL6xgKfeFLg6Q?e=Cs02yZ

Completely crashes Rhino for me. I am opening it whilst on 

https://github.com/BHoM/BHoM/pull/1075 and
https://github.com/BHoM/BHoM_Engine/pull/2160

together with some of the adapter PRs.

This includes some versioning of some create methods. Not sure if the versioning is causing the issue or not (have not tried opening the script on this branch with everything else on master yet).

Opening the script from master on grasshopper and on the PR branches for the others works.

adecler:
@IsakNaslundBh , The crash was caused by a bug in the Rhinoceros toolkit. I have fixed in here: https://github.com/BHoM/Rhinoceros_Toolkit/pull/178

Note that this PR is also depending on https://github.com/BHoM/BHoM_Engine/pull/2163. In your example, you are using a different branch for the BHoM Engine. That PR in the Engine could be merged prior to this one though if you need to use a different branch for the Engine.

al-fisher (dismissed):
@adecler 
Building these three branches against today's alpha I am getting the following error with surfaces:
image

image


A simple script triggering the error is saved here:
https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FGrasshopper%5FToolkit%2FPR567&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4

This works as expected on the alpha

adecler:
Thanks @al-fisher ,

The pre-existing `FromGoo` set of methods had a separate case for `GH_Surface` that was actually not casting to BHoM geometry. This was hidden until now because the old BHoM Goos were only requiring object while now they are more restrictive (e.g. `BH_IBHoMGeometry` actually expects a `IGeometry`). 

Pretty nice that those old errors are finally coming to the surface 😄

IsakNaslundBh (changes requested):
As discussed on call:

script is no longer crashing on startup.

Rectangles still not working, which they were before, so good to get that fixed!

IsakNaslundBh (changes requested):
Rectangles work as expected.

Issue I have found now is with correct error messaging on the component, and for them to not crash.

Some issues found with this:

1. The "ClearAllEvents" in here:

https://github.com/BHoM/Grasshopper_Toolkit/blob/3e2ad90cf7d27d4aa4a2bb1a5920362354789020/Grasshopper_UI/Templates/BakeableParam.cs#L41-L59

Is blocking the error messages produced by grasshopper from being displayed correctly when it fails to convert. This makes the component stay grey and not display that if failed to convert for example an int to a BHoMObject. To be clear, providing an int to a BHoMParameter I think should result in a null, but the component needs to go red and display the error message.

2. Some of the casting is fully crashing out, leading to a popup message stating this. This stops the parameter from executing from that point. It should fail more gracefully, I think by returning `false` from the convert methods when it fail do convert, not by throwing exceptions.

https://burohappold.sharepoint.com/:f:/s/BHoM/EtA2LAi1GhdBpKMHQN8xLEEBqgIVAoMMPThfO_7W6ner8A?e=0scobc

I have put a testscript here called "Grasshopper Parameter Tests".

image


What I would expect in there is:

- No pop up message exceptions to be thrown
- All parameters to contain the same number of values as inputted (nulls for failures)
- All nulls to display a error message of why they could not be converted (Fixing the Clear all events, and making sure the `ConvertFrom` and `ConvertTo` methods return false when they are unable to cast proeprly should take care of this I think)

I also put a script in the same folder with some geometry converts being tested. It does not cover all possible types, but it is a start at least.

Rectangles as well as all surface types I have tested so far works as expected now.

al-fisher (commented):
Thanks @adecler - retested and surface issue is now fixed as per @IsakNaslundBh's note.

Tested @IsakNaslundBh's test scripts and getting same behaviour - description of expected behaviour look good. Thanks for that @IsakNaslundBh  

cheers

adecler:
Thanks @IsakNaslundBh , 

I have corrected the code based on your feedback. I have also tested `GH_IObject`, `GH_IGeometry`, and `GH_Variable` on all the possible type of geometry I could create in Grasshopper and all seems to be working fine. I have updated your test file to reflect that.

After further playing with those goos, I have made a few additional corrections to fix the inconsistencies I could still notice. 

Among those, the one worth noting is how `GH_Variable` handles Rhino geometries. 
- When create something like a CustomObject, the conversion would happen only if the input type was specified as an IGeometry. While I liked the idea that we could still keep the geometry in its Rhino form for some weird edge cases, I didn't like the fact that it was the default behaviour (type hint is System.Object by default so no conversion would apply). 
- The floating Variable param is weird and doesn't even call any of our conversion method. Not sure why but that case is not an issue in my opinion since it will be converted correctly as soo as it is plugged into a BHoM component (even if it's a GH_Variable input). So I wouldn't loose too much sleep over that one.

IsakNaslundBh (commented):
All parameters are working as expected!

Also tested that the variable casting worked as intended after the latest migration, which it seemed to do from the tests I made.

Tested some of the versioning scripts in here:

https://burohappold.sharepoint.com/:f:/s/BHoM/EopxQb5d9QdDqKoD--1178kBUpVHnKPZ1WWwyI2Eg1l18Q?e=sBWLiC

The adapter versioning in the 3.0 folder is struggeling heavily. Not sure if related to this PR. (would expect at least the push and pull to version correctly from the script in there)

Some of the Concrete section creates in 3.2 folder are also struggling. Think that is coming from the fact that the profiles has been upgraded this milestone (changed namespace) while the create methods themselves had a previous attribute on them since before.
Would I need to add another PreviousVersion attribute for this milestone for them?


Also, not sure if the versioning issues have anything to do with this PR or not, but just highlighting here. If not, I can approve if an issue to fix can be raised! 👍

adecler:
> Also, not sure if the versioning issues have anything to do with this PR or not, but just highlighting here. If not, I can approve if an issue to fix can be raised! 👍

Have you compared it to the master ? I don't think versioning shouldn't have anything to do with this but, of course, I could be wrong.

al-fisher (approved):
This is looking like it is casting as expected for core objects now.

I have added an additional section into the Geometry Converts Test - dealing specifically with Rhino Solid representations. 
They are being converted to BHoM Solids correctly as expected.
(GH handles spheres/cones etc. as surfaces and so being cast as Nurbs - which is also as expected)
image

Happy to approve this to get into alphas - if @IsakNaslundBh is happy requested changes are addressed

IsakNaslundBh (approved):
As the only issue I have found does not seem to be related to this PR, I am happy to approve as well.

PR #576, opened 2020-12-08
Grasshopper_Toolkit: Fix explicit casting not working

### LINKED ISSUE #575: Grasshopper_Toolkit: Casting implemented in the oM between different base types does not work
#### Description:

Think the following is critical to get fixed for the beta release, at least to get it back to the previous behaviour, even if it means losing restrictions on parameters/goos.

Can no longer cast from for example geometry to a structural Node:

image

Think this happened when the Goos were limited to only a specific type, rather than `System.Object` in https://github.com/BHoM/Grasshopper_Toolkit/pull/567

Casting within the same Goo category works fine:

image

Not sure if this is at all related to #574 

#### Steps to reproduce:

Try providing a point to something that expects a structural node, or any other casting between for example a geometry and a BHoMObject.

#### Expected behaviour:

Ideal would be to be able to keep that restriction without losing the casting abilities from the oM, but if not possible to fix for the beta, would roll back the restrictions to be just `System.Object` for the beta release, until it can be resolved fully.

#### Test file(s):

https://burohappold.sharepoint.com/:u:/s/BHoM/EVBqRJ4GjNVKtdNDGMe2GTkBl8yGNOOii1Hg4X7UCVFEFw?e=TOsAUB

adecler:

Closes #575

Explicit casting is not working because the casting is to `BHoMObject` and not to the actual type of the expected input. Technically, it would still be possible to solve this **_IF_** the BHoMGoo had access to the expected type. The Param itself holds that information but the relation between the Param and the Goo is all handled internally by Grasshopper so no easy way to get the expected type from the viewpoint of the Goo.

At this point, I think it is sensible to just stop trying to enforce the restrictions on the parameters. The time spent to solve this and the potential resulting bugs are just to big to ignore. SO I reverted back to having `GH_BHoMObject` and `GH_IObject` expecting an object. I have kept `GH_IBHoMGeometry` to still require a `IGeometry` but happy to change that one too if you feel explicit casting is needed there too (note that, as of today, the explicit casting in the geometry oM are all casting from other BHoM geometry).

### Test files
 I have reran the param test script from https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/Grasshopper_Toolkit/_Macro/Grasshopper%20Parameter%20Tests.gh?csf=1&web=1&e=ATRWiG
and all seems to be working fine. The only difference is that the `BHoMObject` and `IObject` parameters are not complaining anymore when provided an invalid type. Which is to be expected as that type might actually end up being valid down the road so not much we can do there.

I have also tested on a couple of object that require an explicit casting :

image

### DIFF
--- a/Grasshopper_UI/Goos/GH_BHoMObject.cs
+++ b/Grasshopper_UI/Goos/GH_BHoMObject.cs
@@ -36,7 +36,7 @@ using BH.Engine.Serialiser;
 
 namespace BH.UI.Grasshopper.Goos
 {
-    public class GH_BHoMObject : GH_BakeableObject<BHoMObject>
+    public class GH_BHoMObject : GH_BakeableObject<object>
     {
         /*******************************************/
         /**** Properties                        ****/
--- a/Grasshopper_UI/Goos/GH_IObject.cs
+++ b/Grasshopper_UI/Goos/GH_IObject.cs
@@ -36,7 +36,7 @@ using BH.Engine.Serialiser;
 
 namespace BH.UI.Grasshopper.Goos
 {
-    public class GH_IObject : GH_BakeableObject<IObject>
+    public class GH_IObject : GH_BakeableObject<object>
     {
         /*******************************************/
         /**** Properties                        ****/


enarhi:
@adecler just tested this as results were causing an error when fed to a bhom object component for internalization. It is working beautifully now. 
image

adecler:
>@adecler just tested this as results were causing an error when fed to a bhom object component for internalization. It is working beautifully now.

@enarhi , `IResults` are not BHoM objects so the correct behaviour should actually be to get an error. In the long run, I recommend using the `IObject` param to internalise results instead. Glad that what was perceived as a downside of this PR by @IsakNaslundBh and I ends up being a positive for others though 😄

IsakNaslundBh (approved):
The immediate issue of the casting has been resolved. Have not had time to test through all other aspects of the toolkit to make sure this has not caused any other side effect.

Approving on this basis.

enarhi:
@adecler, gotcha, makes sense. Just tested with IObject and is working with internalization as expected. Happy to identify an accidental benefit with my results butchering at any rate 😄

PR #596, opened 2021-01-30
Fix Access type update bug

### LINKED ISSUE #518: Grasshopper_Toolkit: Updating access type when versioning
#### Description:

When versioning a method input from a `T` to a `List<T>`, (with the same name) one gets this error:
```
1. This component failed to run properly. Inputs cannot be collected properly.
GetDataList() can only be called on a parameter with access set to GH_ParamAccess.list
```

Which implies that the inputs are getting updated, but not their access type.

might be related to #411 ? but unsure.
#### Steps to reproduce:

Version any method from a item input to a list input.
Keep the name of the input the same as that keeps the wire connected and the component breaks

#### Expected behaviour:

I would like versioning to work and not brick the components.

#### Test file(s):

Can't really add testfile as I did not merge anything where the versioning fails.

adecler:

Closes #518

If a parameter access type (i.e. item, list, tree) was changed by versioning, 
- the Grasshopper toolkit would correctly attempt to update the param 
- BUT would notify BHoM_UI that a parameter was changed in the middle of the process (more precisely when the links were moved from old param to new one)
- BHoM_UI would then take that info into account to recalculate the accessors based on the old input GH just sent.

Not sure if that is clear enough but the solution should be pretty straightforward to understand either way: Grasshopper should notify BHoM_UI of a change on the component ONLY when the user is making a change (e.g. renaming of a param), not when the component is updated because BHoM_UI asked it to. 

### Test files
You probably all have a script for that already given the screenshots you have made so far. If not, you can use what I did for myself:

#### Fake old code

```c#
public static CustomObject ItemToList(object obj, bool b = true)
{
    CustomObject custom = new CustomObject();
    custom.CustomData["data"] = obj.DeepClone();
    return custom;
}

public static CustomObject ListToItem(List<object> objects, bool b = true)
{
    CustomObject custom = new CustomObject();
    custom.CustomData["data"] = objects.ToList();
    return custom;
}
```

#### Fake new code

```c#
[PreviousVersion("4.1", "BH.Engine.Versioning.Compute.ItemToList(System.Object, System.Boolean)")]
public static CustomObject ItemToList(List<object> obj, bool b = true)
{
    CustomObject custom = new CustomObject();
    custom.CustomData["data"] = obj.ToList();
    return custom;
}

[PreviousVersion("4.1", "BH.Engine.Versioning.Compute.ListToItem(System.Collections.Generic.List<System.Object>, System.Boolean)")]
public static CustomObject ListToItem(object objects, bool b = true)
{
    CustomObject custom = new CustomObject();
    custom.CustomData["data"] = objects.DeepClone();
    return custom;
}
```

Just don't forget to also recompile the Versioning toolkit. If you place that code in a different location than me, make sure to correct the string in `PreviousVersion` accordingly.

### DIFF
--- a/Grasshopper_UI/CallerComponent/OnGHParamChanged.cs
+++ b/Grasshopper_UI/CallerComponent/OnGHParamChanged.cs
@@ -58,10 +58,18 @@ namespace BH.UI.Grasshopper.Templates
 
             // Updating Caller.InputParams based on the new Grasshopper parameter just received
             // We update the InputParams with the new type or name
-            Caller.UpdateInput(e.ParameterIndex, e.Parameter.NickName, e.Parameter.Type(Caller));
+            if (m_NotifyChanges)
+                Caller.UpdateInput(e.ParameterIndex, e.Parameter.NickName, e.Parameter.Type(Caller));
 
             return;
         }
+
+
+        /*******************************************/
+        /**** Private Fields                    ****/
+        /*******************************************/
+
+        protected bool m_NotifyChanges = true;
 
         /*******************************************/
     }
--- a/Grasshopper_UI/CallerComponent/OncallerModified.cs
+++ b/Grasshopper_UI/CallerComponent/OncallerModified.cs
@@ -53,6 +53,8 @@ namespace BH.UI.Grasshopper.Templates
             if (update == null)
                 return;
 
+            m_NotifyChanges = false;
+
             // Update the component details
             UpdateComponentDetails(update.ComponentUpdate);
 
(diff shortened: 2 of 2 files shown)


IsakLarbornBH (approved):
Fantastic! Works very well on my ModelLaundry branch, and that's good enough for me. Doesn't break anything else I had on hand to test with, and I don't see why it would. That's an approve from me.

vbeardow (approved):
Fixes the problem from CFD toolkit - thanks!

FraserGreenroyd:
@BHoMBot check copyright compliance

FraserGreenroyd:
@BHoMBot check installer

PR #598, opened 2021-02-03
Added compute method to change background Colour

### LINKED ISSUE #597: Grasshopper_Engine: change background colour method
#### Description:

Add simple utility to change the background colour of the canvas.

alelom:

Closes #597

Useful for taking screenshots for presentations etc.

### Test files

Just drop the component and try it as below:

"Screenshot mode":
image

"Night mode":
image

"Default":
image

### DIFF
--- a/Grasshopper_Engine/Compute/SetCanvasColour.cs
+++ b/Grasshopper_Engine/Compute/SetCanvasColour.cs
@@ -1,0 +1,64 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2021, the respective contributors. All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


adecler (changes requested):
I could see some value to this for users that don't want to manually edit a config file **_IF_**
- The component changes the canvas style only for the script containing it
- Or the component cannot be saved.

As it is right now, sharing a script containing that component permanently changes (i.e. even after closing and reopening Rhino) the canvas style of anyone who opens that script. Even simply dropping that component on the canvas resets its style permanently with no option to cancel that effect.

With that in mind, I don't think the benefit of not having to edit a config file outweighs the downsides. Especially considering that you probably edit that file once and never touch it again (at least that was the case for me).

PR #607, opened 2021-02-19
Fix SetProperty ignoring type hint for the value input

### LINKED ISSUE #605: SetProperty ignored type hint for the value input
#### Description:

Even when type hint is set to double, we get an error complaining about the type being a string:

image

#### Steps to reproduce:
See examples above

adecler:

Closes #605

For SetProperty, the type of the `value` input is object. Which means that the value was always returned without casting regardless of the hint. This is fine for most cases but was failing when a string needed to be parsed as a double for example. 

This PR fixes that by first looking at the hint.

### Test files
image


adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check installer

alelom (approved):
Works, LGTM.

image

PR #609, opened 2021-02-20
Make sure the component is executed when an input name is changed

### LINKED ISSUE #536: Changing input name of CreateCustom component doesn't trigger a run
#### Description:

image

#### Steps to reproduce:
Just change any input name of an input of CreateCustom component

#### Expected behaviour:
Component should re-run

adecler:

Closes #536

The component will be reran every time an input name is changed. 
A few things to note:

-  It is important that we only re-execute the component when the input name changes as `OnGHParamChanged` is also called when a wire is plugged.
- If you test by connecting a `CreateCustom` into an `Explode` component, don't be surprised that the `Explode` asks to be updated. This is a  necessary protective mechanism to avoid loosing wires when outputs change without the user expecting it. So I find testing this PR is easier by connecting a `CreateCustom` component in a `ToJson` component.
- The component will be re-executed on every keystroke. I would have liked to be able to detect when the input menu is closing to avoid that but I couldn't find a way to access it. This is not ideal but I guess too many updates is still better than none.

### Test files
See issue or comment above.

### DIFF
--- a/Grasshopper_UI/CallerComponent/OnGHParamChanged.cs
+++ b/Grasshopper_UI/CallerComponent/OnGHParamChanged.cs
@@ -59,7 +59,13 @@ namespace BH.UI.Grasshopper.Templates
             // Updating Caller.InputParams based on the new Grasshopper parameter just received
             // We update the InputParams with the new type or name
             if (m_NotifyChanges)
+            {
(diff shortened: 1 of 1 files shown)


adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check installer

alelom (approved):
Works, LGTM.

PR #610, opened 2021-02-20
Fix unwanted conversion from Rhino when expected input type is Rhino

### LINKED ISSUE #502: Grasshopper Goo and `FromRhino()` method: FromRhino() for Sphere fails
#### Description:

While doing https://github.com/BHoM/BHoM_Engine/issues/1694, I tried to simply connect a Rhino/Grasshopper Sphere to the FromRhino component for testing.

This always fails:
image

I tried to follow through the code and all goes well until we get to the `FromGoo` method in Grasshopper – hence Grasshopper_Toolkit is probably the right place for this:
image

The dynamic dispatch generates an Exception:
image

which is confusing to me.

#### Expected behaviour:

The component should return a BH.oM.Geometry.Sphere.

adecler:

Closes #502 

A few rare components are actually expecting a Rhino type. The `FromGoo` method was a bit too eager to convert from Rhino regardless of the expected type. The simple fix in this PR corrects that.

### Test files
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Grasshopper_Toolkit/%23502-FromRhinoBug.gh?csf=1&web=1&e=ZMEIif

The top `FromRhino` is supposed to fail because it is expecting a Rhino sphere but receive a Rhino Brep instead. I appreciate this was the original focus of the issue but this is something to do with the Grasshopper Sphere component, not with the BHoM. The `FromRhino(object)` component works perfectly fine with this spherical Brep (and is showing in the Rhino viewport) though so all is good in the world 😉 .

While checking this, I realised that other `FromRhino` components were not working either (e.g. `FromRhino(Line)`). So this ended up being the focus on this PR.

### DIFF
--- a/Grasshopper_UI/Helpers/FromGoo.cs
+++ b/Grasshopper_UI/Helpers/FromGoo.cs
@@ -62,7 +62,7 @@ namespace BH.UI.Grasshopper
             if (data == null)
                 return default(T);
 
-            if (data.GetType().Namespace.StartsWith("Rhino.Geometry"))
+            if (data.GetType().Namespace.StartsWith("Rhino.Geometry") && !typeof(T).Namespace.StartsWith("Rhino.Geometry"))
                 data = BH.Engine.Rhinoceros.Convert.IFromRhino(data);
 
             // Convert the data to an acceptable format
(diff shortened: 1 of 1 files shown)


adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check installer

alelom (approved):
Thanks, LGTM!

PR #611, opened 2021-02-20
Fix default values not assigned for Enumerables

### LINKED ISSUE #528: Automatically generated Create() method does not assign default values for Reference type properties
#### Steps to reproduce/description.

Select an auto-generated Create() method for an object that has some Reference-type properties whose default value is set like:
```cs
public List<string> SomeList { get; set; } = new List<string>() { "Some default value1", "Some default value 2" };
public bool SomeBoolean { get; set; } = true;
```
in the output object, the value will not be set for `SomeList` (an empty list is returned instead), but it will be `true` for `SomeBoolean` and any other Value type.

An example is the [BH.oM.Diffing.DiffConfig](https://github.com/BHoM/BHoM/blob/391660e365360ee075dbb0b1953a5b845460eeca/Diffing_oM/DiffConfig.cs#L34-L72):
image

where this happens:
image

This creates an inconsistency and some issues:

1. I am forced to store the default values for Reference type properties in a "manual" Create() method. 
In my opinion, we should be consistent and store all default values in the same place. So I am tempted to remove all default values from the oM, but that would cause an even worse situation for point 2.

2. Users of the Diffing need to be aware that instantiating a DiffConfig through the auto-generated Create() do not populate some properties with default values. If the auto-generated DiffConfig is used, it results in a behaviour of the diffing which is counter-intuitive for the users and leads them to think that "something is wrong" with the diffing.

3. Personally, I've always liked to think of the auto-generated Create() method as the "parameterless constructor of the BHoM_UI". However, this inconsistency makes it not so. Happy to change my mind, but we still need to be consistent with where we decide to put default values for all types of properties, and I would like for the auto-generated Create() method to actually populate all default values.

@al-fisher @adecler

adecler:

Closes #528 

Default values of Enumerable properties where not set correctly in Grasshopper. See issue for more details.

### Test files
I used what was provided in the issue. Now working fine for me:

image

### DIFF
--- a/Grasshopper_UI/Helpers/ToGH_Param.cs
+++ b/Grasshopper_UI/Helpers/ToGH_Param.cs
@@ -40,6 +40,8 @@ using BH.Adapter;
 using BH.UI.Grasshopper.Templates;
 using BH.Engine.Grasshopper;
 using Grasshopper.Kernel.Parameters.Hints;
+using Grasshopper.Kernel.Types;
+using Grasshopper.Kernel.Data;
 
 namespace BH.UI.Grasshopper
 {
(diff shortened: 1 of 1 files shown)


adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check installer

alelom (approved):
Extremely useful! Thanks!

PR #621, opened 2021-03-29
Add default colour for BHoMObjects

### LINKED ISSUE #620: Grasshopper_Toolkit: BHoMObejcts are not given a render colour
#### Description:

Default colour for BHoMObejct rendering is not set, leading to non selected objects being rendered with `new Color()` rather than the BHoM pink.
#### Steps to reproduce:

build a BHoMObject and try to look at it without it being selected

#### Expected behaviour:

For BHoMObejcts to be rendered with the correct colour.

IsakNaslundBh:

Closes #620

Makes sure BHoMObejcts also gets the correct default colour

### DIFF
--- a/Grasshopper_UI/Goos/GH_BakeableObject.cs
+++ b/Grasshopper_UI/Goos/GH_BakeableObject.cs
@@ -209,6 +209,7 @@ namespace BH.UI.Grasshopper.Goos
             {
                 m_Geometry = (Value as BHoMObject).IGeometry();
                 m_RhinoGeometry = m_Geometry.IToRhino();
+                m_Color = Color.FromArgb(80, 255, 41, 105);//BHoM pink!
                 return true;
             }
             else if (Value is IGeometry)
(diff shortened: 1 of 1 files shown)


rolyhudson (changes requested):
I have my hand up for introducing this problem 🤦 sorry 🙃 about that.
I think we should set the bhom _Coral_ colour once globally either in this class as
`protected Color m_BHoMCoral = Color.FromArgb(80, 255, 41, 105);`
Or as a helper method that just returns the default colour.

IsakNaslundBh:
>I think we should set the bhom Coral colour once globally either in this class as

Agree on that. Was thinking of it as well, just wanted to make as minor change as possible, but have updated

rolyhudson (approved):
LGTM

FraserGreenroyd (changes requested):
I've added a `do-not-merge` tag as there are discussions surrounding whether or not to beta patch this fix in which expect to be concluded tomorrow morning. If we are beta patching then this will need to be handled slightly differently to a normal PR merge, so am just holding things up to make sure we don't lose an opportunity if we want it. If in the morning it is determined we are not beta patching, then I will dismiss this review.

FraserGreenroyd:
@BHoMBot check installer

PR #630, opened 2021-07-09
Firt tests, adding orange capsule, drawing text and drawing image

IsakNaslundBh:

Testing what can easily be achieved changing the UI for alpha only components.

This is not at all in a final state, just checking what is easily doable.

Did some quick experiments with adding an orange text box on top of the old one (was unfortunately not possible to change the colour of the black one without rewriting the whole Render method). It is the downside of the zoomable UI components (the small + and - you get for adding/removing inputs) gets obfuscated a bit.

Also, had a quick test with adding an image or text, which both seem doable quite easily. The current state does not _look_ that great, but did not worry about that yet.

image

Problem with the zoomable parts mentioned:

image

The colours we can use, if we dont want to create our own set of capsules are quite limited to what is built into grasshopper, but think the orange looks ok.

Also, quite small amount of code to get this down, which is nice :)

Option with using the Error colour palette:

image


IsakNaslundBh:
Closing this experiment PR as real PR https://github.com/BHoM/Grasshopper_Toolkit/pull/632 as now merged

PR #656, opened 2022-05-09
Caching of preview meshes to improve performance

### LINKED ISSUE #655: Caching of preview meshes for surface display
#### Description:

Current display of surface based BHoMObjects and BHoM Geometry works, but can get laggy when trying to preview more than a few number of objects. This can be noticed when, for example, a full building of panels and walls is previewed at the same time.

To help mitigate this, preview meshes corresponding to those objects could be computed and cached at first attempt to preview the meshes. This has the potential to drastically speed up the preview.

Related to this is the on the fly computing of the preview material for these meshes. Also this can be improved by pre-computing a DIsplay material matching the colour to be displayed, and also storing the standard preview material with the BHoM coral colour.

IsakNaslundBh:

Closes #655

Changes the way geometry is rendered from rendering Surface geometry via the specific render method accepting Breps, to instead extract PreviewMeshes that are cached at first preview and then reused to preview.

Also storing preview material rather than only preview colour, as creating those for each frame also slows things down.

This will increase the demand on RAM slightly, but from initial testing with a quite significant amount of obejcts this has not yet proven to be a massive problem.

If it is, we could have a look to instead remove the cached BHoM geometry and Rhino geometry, for them to be extracted on the fly instead, and directly turned to the preview meshes.

### Test files

Two testfiles in the folder, first testing that all of our surface types are still visible, second testing the speed. The speed testfile is able to relatively decently handle 100 000 panels with the code on this PR, while it can not handle a fraction of that on main.

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/Grasshopper_Toolkit/%23656-CachingOfPreviewMeshesToImprovePerformance?csf=1&web=1&e=OV1WJr

### DIFF
--- a/Grasshopper_UI/Goos/GH_BakeableObject.cs
+++ b/Grasshopper_UI/Goos/GH_BakeableObject.cs
@@ -185,7 +185,7 @@ namespace BH.UI.Grasshopper.Goos
 
                 if (m_PreviewMesh == null)
                 {
-                    m_IsMeshPreviewable = false;    //If no mesh could be extracted, set flag as no required to check again for the same geometry
+                    m_IsMeshPreviewable = false;    //If no mesh could be extracted, set flag as not required to check again for the same geometry
                     return;
                 }
             }
--- a/Grasshopper_UI/Goos/GH_BakeableObject.cs
+++ b/Grasshopper_UI/Goos/GH_BakeableObject.cs
@@ -36,6 +36,7 @@ using BH.Engine.Serialiser;
 using BH.Engine.Reflection;
 using System.Drawing;
 using BH.oM.Graphics;
+using Rhino.Render;
 
 namespace BH.UI.Grasshopper.Goos
 {
@@ -174,9 +175,35 @@ namespace BH.UI.Grasshopper.Goos
 
         public virtual void DrawViewportMeshes(GH_PreviewMeshArgs args)
         {
-            if (m_RhinoGeometry != null)
-                Render.IRenderRhinoMeshes(m_RhinoGeometry, args, m_Color);
+            if (!m_IsMeshPreviewable)   //Flagged as not previewable - return
+                return;
+
+            if (m_PreviewMesh == null)  //No preview mesh set yet
+            {
+                //Create and store mesh. Will return values for surface type objects (surfaces, breps, meshes etc)
+                m_PreviewMesh = Render.ICreatePreviewMesh(m_RhinoGeometry, args.MeshingParameters);
+
+                if (m_PreviewMesh == null)
+                {
+                    m_IsMeshPreviewable = false;    //If no mesh could be extracted, set flag as no required to check again for the same geometry
+                    return;
+                }
+            }
+
+            if (m_PreviewMesh != null)
+            {
+                if (m_PreviewMesh.VertexColors.Count > 0)   //If colours are set (RenderMeshes) draw these colours
+                {
+                    args.Pipeline.DrawMeshFalseColors(m_PreviewMesh);
+                }
+                else
+                {
+                    Rhino.Display.DisplayMaterial mat = Render.RenderMaterial(args.Material, m_PreviewMaterial);    //If material is default GH material, BHoM default will be used, if not, default GH material is used.
+                    args.Pipeline.DrawMeshShaded(m_PreviewMesh, mat);
+                }
+            }
         }
+
 
         /***************************************************/
 
@@ -193,8 +220,11 @@ namespace BH.UI.Grasshopper.Goos
 
         private bool SetGeometry()
         {
+            ResetPreviewMeshes();   //Clears cashed preview meshes and resets preview flag.
+
             if (Value == null)
             {
+                m_IsMeshPreviewable = false;
                 return true;
             }
             else if (Value is IRender)
@@ -208,6 +238,8 @@ namespace BH.UI.Grasshopper.Goos
                     m_thickness = (Value as RenderCurve).Thickness;
                     m_Geometry = (Value as RenderCurve).Curve;
                 }
+                double transparency = (255 - m_Color.A) / (double)255;
+                m_PreviewMaterial = new Rhino.Display.DisplayMaterial(m_Color, transparency);
                 return true;
             }
             else if (Value is BHoMObject)
@@ -224,8 +256,22 @@ namespace BH.UI.Grasshopper.Goos
             }
             else
             {
+                m_IsMeshPreviewable = false;
                 return false;
             }
+        }
+
+        /***************************************************/
+
+        protected virtual void ResetPreviewMeshes()
+        {
+            if (m_PreviewMesh != null)
+            {
+                m_PreviewMesh.Dispose();
+                m_PreviewMesh = null;
+            }
+
+            m_IsMeshPreviewable = true; //Default to true until checked
         }
 
         /***************************************************/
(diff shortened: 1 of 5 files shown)


Grasshopper_UI/Goos/GH_BakeableObject.cs
```diff
@@ -174,10 +175,36 @@ public override bool Write(GH_IWriter writer)
...
+                return;
+
+            if (m_PreviewMesh == null)  //No preview mesh set yet
+            {
+                //Create and store mesh. Will return values for surface type objects (surfaces, breps, meshes etc)
+                m_PreviewMesh = Render.ICreatePreviewMesh(m_RhinoGeometry, args.MeshingParameters);
+
+                if (m_PreviewMesh == null)
+                {
+                    m_IsMeshPreviewable = false;    //If no mesh could be extracted, set flag as no required to check again for the same geometry
```
FraserGreenroyd, line 188:
```suggestion
                    m_IsMeshPreviewable = false;    //If no mesh could be extracted, set flag as not required to check again for the same geometry
```

Grasshopper_UI/Render/CreatePreviewMesh.cs
```diff
@@ -0,0 +1,192 @@
...
+        /**** Public Methods  - Interfaces              ****/
+        /***************************************************/
+
+        public static RHG.Mesh ICreatePreviewMesh(this object obj, RHG.MeshingParameters parameters)
+        {
+            if (obj == null)
+            {
+                return null;
+            }
+            if (parameters == null)
```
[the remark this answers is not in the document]
FraserGreenroyd: Or at least a blank line after the brackets if you prefer the use of the brackets, but separation of `if(params == null)` from the rest is good.

Grasshopper_UI/Render/CreatePreviewMesh.cs
```diff
@@ -0,0 +1,192 @@
...
+            RHG.Mesh[] array = RHG.Mesh.CreateFromBrep(brep, parameters);
+            if (array == null)
+            {
+                return new RHG.Mesh();
+            }
+            if (array.Length == 1)
+            {
+                return array[0];
+            }
+            RHG.Mesh mesh = new RHG.Mesh();
```
FraserGreenroyd, line 115:
```suggestion
            if (array == null)
                return new RHG.Mesh();

            if (array.Length == 1)
                return array[0];

            RHG.Mesh mesh = new RHG.Mesh();
```

Again just giving some spacing around the different checks/blocks. Happy to keep the brackets but if so can we then put a blank line between each different `if` block instead.

Grasshopper_UI/Render/RenderMaterial.cs
```diff
@@ -50,6 +50,27 @@ public static DisplayMaterial RenderMaterial(DisplayMaterial material, Color cus
...
+        public static DisplayMaterial RenderMaterial(DisplayMaterial material, DisplayMaterial custom)
+        {
+            if (custom == null)
+                return material;
+
+            Color pColour = GH.Instances.ActiveCanvas.Document.PreviewColour;
+            Color ghColour = material.Diffuse;
+            if (ghColour.R == pColour.R & // If the color sent by PreviewArgs is the default object PreviewColour
+                ghColour.G == pColour.G &
+                ghColour.B == pColour.B) // Excluding Alpha channel from comparison
```
FraserGreenroyd, line 63: Surely if any of these are `!=` then we don't need to check the others? In which case it's better to use `&&` than `&` I'd imagine?

Grasshopper_UI/Goos/GH_BakeableObject.cs
```diff
@@ -193,8 +220,11 @@ public virtual void DrawViewportWires(GH_PreviewWireArgs args)
 
         private bool SetGeometry()
         {
+            ResetPreviewMeshes();   //Clears cashed preview meshes and resets preview flag.
```
FraserGreenroyd, line 223: Just to check, will this be called any time any geometry of any object being rendered is changed/set? As in, if I have a panel, but then change its geometry, will be render with the changed geometry or will it use the cached geometry?

FraserGreenroyd (changes requested):
Some minor spelling/formatting changes requested from code review (functionality review not yet performed).

Tagging @alelom as well to take a glance through just to check if this causes any concern for him with the remote compute work he's working on following our nuget discussion this morning @alelom ?

Happy for conversations to be marked resolved if an answer is given @IsakNaslundBh 👍

Grasshopper_UI/Goos/GH_BakeableObject.cs
IsakNaslundBh: That is a good question, but do not think this PR will change from the current behaviour (where we are using cached Rhino geometry to render anyway).

The SetGeometry method is called when the Value of the Goo is set, https://github.com/BHoM/Grasshopper_Toolkit/blob/d3d16e824d779e7e4b6f68f9fb9458f023978fa4/Grasshopper_UI/Goos/GH_BakeableObject.cs#L58-L69

So that case is handled. I guess, theoretically, you could get the reference of a Panel from a particular `GH_BakeableObject` and change it without this method being called again, and hence preview being wrong. I think that never happens in practice though, and if it did I do not think this PR makes it worse.

Worth to keep an eye on for sure, but think it is a separate issue that we can log if you think required.

Grasshopper_UI/Render/CreatePreviewMesh.cs
```diff
@@ -0,0 +1,192 @@
...
+                return null;
+            }
+            if (parameters == null)
+                parameters = RHG.MeshingParameters.Default;
+
+            try
+            {
+                return CreatePreviewMesh(obj as dynamic, parameters);
+            }
+            catch (Exception)
```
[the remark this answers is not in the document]
IsakNaslundBh: Dont think so. Would prefer for the preview system to give as little errors on components as possible. Also, doubt this will crash that often (if any time at all), but good to have as a safety.

Also, for this, I tried to just follow what had been done in the other files :)

Grasshopper_UI/Render/RenderMaterial.cs
IsakNaslundBh: fully agree. Just copied the other method and did not really think about it. Will fix, and might fix the other method as well.

Grasshopper_UI/Goos/GH_BakeableObject.cs
FraserGreenroyd: Happy with that answer @IsakNaslundBh - happy for it to not be a separate issue at this stage until we find an issue

Grasshopper_UI/Render/CreatePreviewMesh.cs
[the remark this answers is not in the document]
FraserGreenroyd: Understood, sounds good to me 😄

michaelhoehn (approved):
I've compared the code on main and the PR code against the sample scripts and am very happy with the performance improvements. I didn't notice any adverse side effects on memory usage when upping the panel counts quite significantly. The only real difference was noticed when orbiting/panning around the models which provided a significant framerate performance boost. Happy to see this merged 🐛

FraserGreenroyd (approved):
Changes are good, happy with the discussion, great work, thanks @IsakNaslundBh 

@alelom unless you desperately need this for your remote computing work, I won't plan to make a nuget out of this until the next beta - but if you do need it desperately then let me know!

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #676, opened 2023-01-23
Add null-check for request before checking SelectedItem on global search

### LINKED ISSUE #675: Stop crash on global search when request is null
#### Description:

When debugging in Grasshopper, quite often the code crashes (non-fatally) on https://github.com/BHoM/Grasshopper_UI/blob/627df5a6b09fe0bc2bcd526bd41d0535074ad893/Grasshopper_UI/Global/GlobalSearch.cs#L207

due to the request being null.

Need to add a null-check in this method for the request.

Can be slightly annoying for debugging as this can happen when you have any repository attached, as long as you have built Grasshopper yourself.

Should not be critical for production code as the whole segment is in a try-catch block, but null check should be beneficial anyway.

#### Steps to reproduce:

Build Grasshopper_UI
Attach any BHoM solution to Grasshopper (does _not_ have to be Grasshopper)
Use grasshopper, dragging some wires etc. 
For me it relatively frequently crashes out on this line.

#### Expected behaviour:

For the null-request to be checked for and handled

#### Test file(s):

Happens on all files for me, no test file hence required.

IsakNaslundBh:

Closes #675

Adding simple null-check on request before trying to access SelectedItem.
See issue as to why/when this has been slightly annoying issue.

### Test files

Global search should still work.

### Additional comments

Cannot really see how this could have a negative impact on anything, but glad for a quick look @adecler

### DIFF
--- a/Grasshopper_UI/Global/GlobalSearch.cs
+++ b/Grasshopper_UI/Global/GlobalSearch.cs
@@ -204,7 +204,7 @@ namespace BH.UI.Grasshopper.Global
 
                     componentCreated = canvas.InstantiateNewObject(node.Id, initCode, canvas.CursorCanvasPosition, true);
                 }
-                else if (request.SelectedItem is CustomItem)
+                else if (request?.SelectedItem is CustomItem)
                 {
                     CustomItem item = request.SelectedItem as CustomItem;
                     if (item != null && item.Content is IGH_ObjectProxy)


adecler (approved):
I don't see any reason why a null check would be a problem here. Approved!

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd (approved):
As per @adecler - can't see this hurting anything. But I look forward to being proved wrong in 3 weeks time...

PR #710, opened 2024-10-04
Add BoxMapping for display mesh

### LINKED ISSUE #709: Add texture mapping to Meshes
#### Description:

Currently if a [RenderGeometry ](https://github.com/BHoM/BHoM/blob/develop/Graphics_oM/Render/RenderGeometry.cs)is being generated and the `SurfaceColour` Texture is assigned a bitmap texture _and_ the geometry of the `RenderGeometry` is a mesh, no texture is displayed on the object. After some digging, found that this is due to the mesh converted from BHoM is lacking `TextureCoordinates` for how to map the texture to the mesh. 

Need to add a call to assign those texture coordinates for the display mesh in rhino. A simple box mapping in world coordinates I think is a good enough place to start for this general case, and if we in the future want more control over this, I think we need to add a new class in the Graphics_oM  with more properties able to handle this.

as a note, this is only an issue for RenderGeometries with Mesh as geometry. If assigning a surface, the texture mapping is handled automatically by rhino when generating the display mesh.

IsakNaslundBh:

Closes #709

Adds box mapping to display mesh. This ensures textures are displayed on meshes - in particular critical for when a RenderGeometry with an assigned Texture with a bitmap is returned.

This could potentially be improved/expanded upon by changing the texture mapping method, but this simple box mapping seem to work decently well and be a good first step.

I tried also to use the `CreateCustomMeshMapping`, providing the mesh, but that gave worse results, in particular for vertical faces that got the texture displayed as stripes of the value applicable to a particular x,y coordinate.


IsakNaslundBh:
@BHoMBot check versioning

michaelhoehn (approved):
Happy with this update, thanks Isak! 👍

michaelhoehn:
@BHoMBot check ready-to-merge

PR #733, opened 2026-02-10
Improve BHoM loading time in Grasshopper

### LINKED ISSUE #732: Slow UI loading time due to assembly folder scanning
#### Description:
When the BHoM_UI is first loaded by Grasshopper, the entire Assembly folder is being scanned. This results in an additional delay of over 60 seconds in the loading time of the UI. It would be good to find a solution that eliminate that initial delay, at least partially.

adecler:
### NOTE: Relates to
https://github.com/BHoM/BHoM_UI/pull/531

 ### NOTE: Depends on 
https://github.com/BHoM/BHoM_Installer/pull/253
  
Closes #732 

This PR addresses https://github.com/BHoM/BHoM_UI/issues/521 alongside the BHoM_UI PR linked above.

This specifically addresses the slow loading time of BHoM_UI in Grasshopper despite our lazy loading strategy for the assemblies.

This was caused by a full scan of our assembly folder as soon as Grasshopper was trying to load our UI. With over 600 dlls, this was taking over a minute with no benefit to our strategy.

Based on a test on my machines, loading the BHoM_UI on a fresh machine reboot gives us this improvement on the folder scanning part of the UI loading:
 - Without this PR: **68s** 
 - With this PR: **3.7s** 

The solution implemented here is to move the dlls directly required by our Grasshopper plugin to a new `C:\ProgramData\BHoM\GrasshopperPlugin" folder.

**Note**: This PR is enough for local testing but we will need to also fix both installers before this PR can be merged.
**Note 2**: This PR can be merged after the BHoM_UI PR is merged as it solves a separate problem.

### Test files

Continue your testing from https://github.com/BHoM/BHoM_UI/pull/531 but with this code compiled.

### DIFF
--- a/Grasshopper_Engine/Grasshopper_Engine.csproj
+++ b/Grasshopper_Engine/Grasshopper_Engine.csproj
@@ -20,62 +20,62 @@
     <Reference Include="BHoM">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>$(ProgramData)\BHoM\Assemblies\BHoM.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
     </Reference>
     <Reference Include="BHoM_Engine">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>$(ProgramData)\BHoM\Assemblies\BHoM_Engine.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
     </Reference>
     <Reference Include="BHoM_UI">
       <HintPath>$(ProgramData)\BHoM\Assemblies\BHoM_UI.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
       <SpecificVersion>False</SpecificVersion>
     </Reference>
     <Reference Include="Data_Engine">
       <HintPath>$(ProgramData)\BHoM\Assemblies\Data_Engine.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
       <SpecificVersion>False</SpecificVersion>
     </Reference>
     <Reference Include="Dimensional_oM">
       <HintPath>$(ProgramData)\BHoM\Assemblies\Dimensional_oM.dll</HintPath>
-      <Private>False</Private>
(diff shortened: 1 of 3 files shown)


michaelhoehn (dismissed):
I have tested successfully and experienced the speed improvements ⏩ 😄
Having tested from a custom BHoM build (compiled locally), I'm happy to approve based on the post-build event logic working correctly on clean installs. I couldn't test the installer PR without a testing procedure, but the code appears sound.

Before merging, could you @adecler please confirm two things:

1. Is the Grasshopper.dll copy intentional? The post-build event copies all DLLs (*.dll) to the GrasshopperPlugin folder, including Grasshopper.dll itself. In my testing, this causes a duplication error on Grasshopper startup (see screenshot below). I had to manually delete this file to proceed. Should we be filtering this out in the xcopy command, or is there a reason we need our own copy of it?
2. Does the installer handle legacy folder cleanup? My understanding is that our installer does a full wipe of ProgramData\BHoM on install, which would automatically handle this - but I want to confirm before merging, since the Grasshopper duplication error is quite alarming for users who don't know what they're looking at.

Image

adecler:
@michaelhoehn , I don't have a dll called `Grasshopper.dll` in my Build folder nor in `C:\ProgramData\BHoM\GrasshopperPlugin`. Maybe this is a legacy dll? Can you delete your build folder and try again ?

Looking into this, however, showed me that we could have an error on the post-build event when Grasshopper had already been installed so I fixed that in my last commit (https://github.com/BHoM/Grasshopper_UI/pull/733/commits/7987f295b7c9c9759a1204cb7319b0f630dfa0e2)

michaelhoehn (approved):
I tested again after fully deleting any previous build folder and the new proposed folder so I had a completely clean testing environment. After testing again I did not see the error again, which suggests the dll duplication was due to code on my machine during testing. I'm happy with the performance and approach here.

adecler:
@BHoMBot check ready-to-merge

adecler:
@BHoMBot check ready-to-merge

PR #745, opened 2026-04-29
Make sure the wire menu only shows when a wire is dragged

### LINKED ISSUE #743: Pop-up menu on components connectors prevents the creation of relays
#### Description:
Normally, relays can be created by double-clicking on inputs and outputs connectors but the BHoM menu prevents that from happening.

This relates to this issue: https://github.com/BHoM/Grasshopper_UI/issues/593
The conversation in that issue leads to the creation of settings to let the user control the behaviour of menus. This addresses the narrower scope of fixing the immediate problem of the menu showing up by clicking on connectors. 

#### Steps to reproduce:
Try to double-click on an input or output connector.

#### Expected behaviour:
The menu should only show up when a wire is being dragged, not when a connector is clicked.

#### Test file(s):
Test on any component

adecler:

Closes #743 

As described in the issue, the BHoM wire menu was also showing up with a simple click on a component connector, without the need to actually drag a wire. This was preventing other GH functionalities like the creation of relays by double-clicking on a connector.  

This PR fixes that by a travel distance threshold before the menu is triggered on mouse-up events.

This PR also limits the wire menu to BHoM components. The menu is irrelevant for non-BHoM components since it will not be able to filter much and can potentially be annoying when people are not using BHoM components, even with the bug fixed. See this issue for more details: https://github.com/BHoM/Grasshopper_UI/issues/593

### Test files
No need for a test file here, 
 - Just create any component and interact with its connectors in any way you can think of and check that the menu only pops up when it makes sense. 
- Create various types of BHoM components (param components, dropdown components, ...) and make sure the wire menu works on them. Then create non-BHoM components and make sure the wire menu does show up on them.

### DIFF
--- a/Grasshopper_UI/Global/GlobalSearch.cs
+++ b/Grasshopper_UI/Global/GlobalSearch.cs
@@ -164,18 +164,25 @@ namespace BH.UI.Grasshopper.Global
             FieldInfo targetField = typeof(GH_WireInteraction).GetField("m_target", BindingFlags.NonPublic | BindingFlags.Instance);
             if (targetField != null && targetField.GetValue(m_LastWire.Wire) != null)
             {
-                m_LastWire = null;
                 Debug.WriteLine(DateTime.Now.ToString("HH:mm:ss") + "- Wire info cleared.");
             }
-            else
+            else if (m_LastWire.Wire.ControlPointDown != null)
             {
-                GlobalSearch.Open(canvas.FindForm(), new SearchConfig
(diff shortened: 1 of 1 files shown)


michaelhoehn (approved):
Works as described. Happy to go for merge on this one. Thanks!

Tested in both Rhino 7 and 8 on both BHoM and non-BHoM components as recommended.

adecler:
@BHoMBot check ready-to-merge

PR #747, opened 2026-05-03
Fix dictionary input set to null after component copy

adecler:
Closes #746 

Default value for dictionary inputs were not set correctly when a component was copied. 
The same was happening if the dictionary input had been internalized. When the component was copied, the internalised input would be lost.

### Test files
Make sure copied components maintain their dictionary input both for default value and internalised value. 
Here's a script that can help with that:
[DictionariesSetToNullWhenCopied.zip](https://github.com/user-attachments/files/27319196/DictionariesSetToNullWhenCopied.zip)


michaelhoehn (approved):
Functions as expected. Successfully tested and approved 👍

adecler:
@BHoMBot check ready-to-merge

PR #748, opened 2026-05-04
Making sure that empty branch inputs produce empty branch outputs

### LINKED ISSUE #701: Empty lists get turned into nulls on object creation
#### Description:

Not entirely sure if the right repo to raise an issue, but it seems to be a UI glitch to me. When creating an object using property assignment, empty list inputs get converted to `null` - in the example below we should rather get 3 valid objects and ignore 2 empty lists rather than getting 5 objects, 2 of them invalid:
image

#### Test file(s):

On [SharePoint]()

adecler:

Closes #701

This creates the behaviour requested in the issue.

That being said, I am not a big fan of this PR for a few reasons:
- The code needed to get this behaviour doesn't fell great. We're effectively having to access each input three times (once in `SolveInstance`, once when actually executing the component, and once in `AfterSolveInstance`. The code also has a set of guardrails to make sure that behaviour is only occurring when for that specific scenario but it would be easy to miss another edge-case where it is now happening but is not desired. I am not fluent enough in complex GH workflow to tell.
- This fix is specific to single-item inputs (i.e. each item is producing an output vs objects with lists as property). Changing the behaviour for this case raises questions on the expected behaviour for list and tree inputs. For example, an empty branch is treated as an empty list and therefore produces an output if the component was expecting a list for that input. That doesn't align completely with the change we're making here and could confuse the user in other ways.  
- There a lot of lacing situations that are not covered here that could be problematic. For example, what if all inputs have values for the same corresponding branch but a single input has that branch empty? Does the use expect to still produce an object? If yes, which what value for that input? Null? Default value? We're just opening the door to more confusing rules.

So, in summary, it feels that this PR is creating more problems than it is solving. It feels it would be preferable for everyone to have simple expected behaviour on BHoM components and, in exchange, ask users to keep their script clean. But I am not a GH expect so maybe I am  missing something?

@michaelhoehn , I would appreciate your opinion on this one.

### Test files
See issue

### DIFF
--- a/Grasshopper_UI/CallerComponent/GH_Component.cs
+++ b/Grasshopper_UI/CallerComponent/GH_Component.cs
@@ -20,27 +20,29 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
-using System;
-using System.Linq;
-using GH = Grasshopper;
-using Grasshopper.Kernel;
-using BH.oM.Base;
-using BH.oM.UI;
-using System.Collections.Generic;
-using System.Windows.Forms;
-using BH.UI.Grasshopper.Global;
-using Grasshopper.Kernel.Parameters;
-using BH.UI.Grasshopper.Parameters;
-using BH.Engine.Reflection;
-using BH.oM.Geometry;
-using BH.Engine.Grasshopper;
-using BH.UI.Grasshopper.Components;
-using System.Collections;
 using BH.Adapter;
+using BH.Engine.Grasshopper;
+using BH.Engine.Reflection;
+using BH.oM.Base;
 using BH.oM.Base.Debugging;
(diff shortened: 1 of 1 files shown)


michaelhoehn:
I have reviewed the functionality and the code with @adecler. Although the code does provide a fix to the proposed issue, I personally feel it responds to something that becomes outside of the scope of the BHoM in general. What could be solved by keeping data trees tidy as a first principle of visual scripting in Grasshopper should not, in my view, be imposed or alter the general logic of the BHoM grasshopper UI to make things tidy for the users. I say this because while a few may want this behavior, not everyone would expect or appreciate this functionality from the BHoM. 

At a minimum I would table this and recommend a deeper polling of the user base to study and respond to a broader group of users before altering this behavior. I will be closing this but please feel free to reopen in the future should more demand arise.

